PR #52, opened 2016-03-30
Support for universal apps / server-side rendering

loudwinston:
### CHANGE SUMMARY
- Add `loadNamespaces` method that iterates through components, discovering and loading namespaces that are declared using `@translate`.  Works very similarly to `redux-async-connect`.  Designed be used with `react-router` `match` method.
- Add `namespaces` property to `Translate` wrapper returned by `@translate`.


jamuhl:
excellent...will have a look later and merge / publish asap. thanks a lot...

jamuhl:
in react-i18next@1.3.0...thanks again

loudwinston:
No problem @jamuhl !

PR #71, opened 2016-04-21
Support empty translate hoc with defaultNS

perrin4869:
Somewhat self explanatory. In my app I have one single namespace right now, so I would like to use it by default without passing it to the HoC each time.


jamuhl:
published....react-i18next@1.3.1

perrin4869:
Thanks!!!

PR #77, opened 2016-04-23
Add express middleware integration explanation

perrin4869:
Adapted from the [sample/boilerplate](https://github.com/simpleblack/react-redux-universal-hot-example) example, thought this might be useful to add to the `README` since it's not trivial how to use the express middleware together with `react-i18next`.
Not sure if this is the best solution, but it does the job well enough!

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,8 +132,27 @@ import { match } from 'react-router';
 match({...matchArguments}, (error, redirectLocation, renderProps) => {
    loadNamespaces({ ...renderProps, i18n: i18nInstance })
    .then(()=>{
-		// All i18n namespaces required to render this route are loaded   
+      // All i18n namespaces required to render this route are loaded   
    })
+});
+```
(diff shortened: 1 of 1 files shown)


jamuhl:
why not using req.i18n ?

perrin4869:
It doesn't work because `req.i18n` is not an `i18next` instance, it's just an object literal:

https://github.com/i18next/i18next-express-middleware/blob/master/src/index.js#L41

Therefore, you cannot pass it to `I18nextProvider`, which requires a complete `i18next` instance.

jamuhl:
right....

perrin4869:
Maybe it would be a good idea to turn `req.i18n` into an i18next instance, although that might not be trivial to do or might require a breaking change, especially with the `changeLanguage` method.

jamuhl:
yes...might cause some trouble - so not sure.

perrin4869:
how expensive of a method is `i18next.cloneInstance`? As long as it's not very demanding I think this is good enough.

jamuhl:
not to expensive i think...but we might need to try and find out

PR #78, opened 2016-04-25
Update docs for i18next-express-middleware usage

perrin4869:


jamuhl:
thank you for updating...

PR #82, opened 2016-04-28
Add getWrappedInstance() to translate wrapper

vpriem:
Sometimes you need to access to the component instance wrapped in the component returned by translate().

No magic, just copied the approach of react-redux [connect](https://github.com/reactjs/react-redux/blob/master/docs/api.md#connectmapstatetoprops-mapdispatchtoprops-mergeprops-options).

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,11 +46,12 @@ ReactDOM.render(
 
 ### Translate HOC
 
-__translate(namespaces)__: higher-order component to wrap a translatable component.
+__translate(namespaces, options)__: higher-order component to wrap a translatable component.
 
 - All given namespaces will be loaded.
 - props.t will default to first namespace in array of given namespaces (providing a string as namespace will convert automatically to array, providing no namespaces will default to `defaultNS`)
 - used nested inside I18nextProvider (context.i18n)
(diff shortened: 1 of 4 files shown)


jamuhl:
why not just https://github.com/i18next/react-i18next/blob/master/src/translate.js#L65

vpriem:
As far as i understand `Translate.WrappedComponent` is more to access to the component, for server side rendering as described in pull request [#15](https://github.com/i18next/react-i18next/pull/15), but not to the rendered instance itself. With `getWrappedInstance()` you could access to the instance using `ref`. Should I provide an example?

jamuhl:
ok i see. can you remove the dependency on invariant? prefer to keep dependencies low.

jamuhl:
will be merged afterwards...

vpriem:
I was thinking of that. Of course I can but should I remove invariant completely, throwing an exception if `{ withRef: true }` was not provided? Or should I add `invariant` as external to the rollup config and let developers manage the dependency by themselves?

jamuhl:
i would prefer removing it completely...console.error might be good enough?

vpriem:
As you want, just removed all invariant deps and added `console.error`.

jamuhl:
published react-i18next@1.4.0

Issue #83, opened 2016-04-29
WithRef


duvillierA:
Hello,

Can you provide a clean example about using `{WithRef: true}` and `getWrappedInstance()` to get refs inside the parent container?

Thanks.

jamuhl:
@vpriem could you help out here?

duvillierA:
thanks @jamuhl 

Looking forward a response

PR #85, opened 2016-05-02
getWrappedInstance() clean example

vpriem:
Following issue https://github.com/i18next/react-i18next/issues/83


README.md
```diff
@@ -73,6 +73,54 @@ export default translate(['defaultNamespace', 'anotherNamespace'])(TranslatableV
...
+```
+
+```javascript
+import React, { Component } from 'react';
+import ./TranslatableView;
+
+class App extends Component {
+
+  handleClick() {
+    this.refs.translatedView.foo();
```
piotr-cz: Shouldn't this line be `this.refs.translatedView.getWrappedInstance().foo()` ?

jamuhl: @piotr-cz yes...agree...that readme was changed meanwhile and the new docu does not contain that sample anymore: https://react.i18next.com/components/translate-hoc.html

Issue #6, opened 2016-01-22
Incompatible with hot module reload?


Danita:
First of all, thank you for this work, :smile: it really looks a lot easier to use than `react-intl` which I find a little overkill for a simple app.

I have a translated view that goes like this:

```jsx
import React from 'react';
import { translate } from 'react-i18next/lib';

const LoginForm = React.createClass({
	render() {
		const t = this.props.t;
		console.log(t); // <-- we should have the translation fn here
		return (
			<div>
				<h3>{ t('Sign in')}</h3>
				{/*...*/}
			</div>
		);
	}
});

export default translate(['login'])(LoginForm);
```
On first load and manual reload it works as advertised, displaying the translated message. The `console.log(t)` call returns something like `function fixedT(key, options) {...}`, but when I modify something on the component and thus triggering the hot reload module (Webpack), I get `undefined` and an `Uncaught TypeError: t is not a function`.

Is there any way to make react-i18next compatible with HMR?

Thanks!

jamuhl:
will have a look on monday.

jamuhl:
tried to reproduce on our env...but not having an issue with HMR...but might be we got a different configuration plus different versions of webpack.

Do you have a small project i could access to reproduce?

Danita:
Hey Jan, thanks for taking a look at this. I have the latest versions of all dependencies, including Webpack. As of now, I replaced `react-i18next` for `react-intl`, but I think I still have the branch where I tried `react-18next`, so I could upload a repro case.

jamuhl:
never mind...thanks for the reply...closing this for now.

sorahn:
I just ran into this same issue.

I'm using the code as a decorator

```javascript
@translate(['common'])
class Header extends Component {
  static propTypes = {
    t: PropTypes.func
  };

  render () {
    const { t } = this.props

    return <h1>{t('hello_world')}</h1>
  }
}
```

Here's my i18n.js init.

```javascript
import i18n from 'i18next'
import languages from './strings'

i18n
  .init({
    lng: 'en', // set dynamically on build
    // lngs: Object.keys(languages),
    fallbackLng: 'en',

    ns: ['common'],
    defaultNS: 'common',

    debug: true,

    interpolation: {
      escapeValue: false // not needed for react!!
    }
  })

Object.keys(languages).map(lang => {
  i18n.addResourceBundle(lang, 'common', languages[lang], true)
})

export default i18n
```

strings/index.js
```javascript
import en from './en.json'
[...]

PR #112, opened 2016-06-06
Fix t is not defined when hot reload

roadhump:
It looks like one issue from here https://github.com/i18next/react-i18next/issues/6

```
export const Menu = ({t}) => (
    <div>{t('hello')}</div>
);

export default translate(['menu'])(Menu);
```

When I use HMR, this code will throw `t is not defined` on reload. It seems to happen because `t` is a local var of `translate` decorator and it is not preserved between reloads. So I set it as class property and it works as expected.

### DIFF
--- a/src/translate.js
+++ b/src/translate.js
@@ -8,7 +8,7 @@ export default function translate(namespaces, options = {}) {
   const { withRef = false } = options;
 
   return function Wrapper(WrappedComponent) {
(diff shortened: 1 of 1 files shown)


jamuhl:
published react-i18next@1.4.2

wong2:
thanks, this solved my problem timely

Issue #62, opened 2016-04-11
Cannot read property 'loadNamespaces' of undefined


rpnna:
Hi, I run into this issue:
```javascript
Uncaught TypeError: Cannot read property 'loadNamespaces' of undefined
i18next::backendConnector: loaded namespace common for language en-US Object {appName: "app name sample", interpolateSample: "you can interplate {{value}} or {{component}} via interpolate component!"}
i18next: languageChanged en-US
i18next: initialized
Object {debug: true, ns: Array[1], defaultNS: "common", fallbackLng: Array[1], fallbackNS: false…}
```
app.js is the following:
```jsx
class App extends Component {

    ...

    render () {
        return (
            <Router history={hashHistory}>
                <Route path='/' component={Homepage}>
                    <Route path='signin' component={Signin} />
                    <Route path='signup' component={Signup} />
                    ...
                    ...
                </Route>
            </Router>
        );
    }
}


const RootedApp = root(App, store);
render(<I18nextProvider i18n={i18n}><RootedApp /></I18nextProvider >, document.getElementById('app'));
```

i18n.js
```jsx
import i18n from 'i18next';
import XHR from 'i18next-xhr-backend';
import LanguageDetector from 'i18next-browser-languagedetector';

i18n
.use(XHR)
.use(LanguageDetector)
.init({
    fallbackLng: 'en',

    ns: ['common'],
    defaultNS: 'common',

    debug: true,

    interpolation: {
        escapeValue: false
    }
});

export default i18n;
```

Then, in my Home.js:
```jsx
import React, {Component} from 'react';
import { translate } from 'react-i18next';

@translate(['view', 'nav'])
class Home extends Component {

    render () {
        const { t } = this.props;

        return (
            <div'>
                <h1>{t('common:appName')}</h1>
                {this.props.children}
            </div>
        );
    }
};

```
Anyway, i18next seems to be correctly loaded. What's the problem?

Thanks

jamuhl:
Seems like i18n gets not set on context in your case...which is really odd. Does the error show up in browser or test environment?

rpnna:
Hey, thanks for collaboration.
The error occurs in the browser..

jamuhl:
```
const RootedApp = root(App, store);
render(<I18nextProvider i18n={i18n}><RootedApp /></I18nextProvider >, document.getElementById('app'));
```

[...]

PR #131, opened 2016-06-17
Fix i18n is not defined on component mount/unmount after hmr

TLadd:
I've been experiencing errors with hot module reloading with react-i18next. https://github.com/i18next/react-i18next/pull/112 helped; I don't get an error immediately on the hot module reload anymore.

However, I still do get an error when a hot module reload occurs, and I then do something which mounts or unmounts a translated component (route change, for instance).

Error is either
- Cannot read property 'loadNamespaces' of undefined (https://github.com/i18next/react-i18next/issues/62)
- Cannot read property 'off' of undefined

depending on whether componentWillMount or componentWillUnmount occurs first.

Cause and fix are the same as https://github.com/i18next/react-i18next/pull/112, just with the i18n local variable this time.

### DIFF
--- a/src/translate.js
+++ b/src/translate.js
@@ -8,13 +8,12 @@ export default function translate(namespaces, options = {}) {
   const { withRef = false, wait = false } = options;
 
   return function Wrapper(WrappedComponent) {
-    let i18n;
 
     class Translate extends Component {
       constructor(props, context) {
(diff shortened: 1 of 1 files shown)


jamuhl:
just published react-i18next@1.5.1

thanks for your PR.

PR #151, opened 2016-07-05
Hoist non react statics

perrin4869:
Thisi is what [react-redux](https://github.com/reactjs/react-redux) `connect`, and other `HoC` do.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,9 @@
     "type": "git",
     "url": "https://github.com/i18next/react-i18next.git"
   },
-  "dependencies": {},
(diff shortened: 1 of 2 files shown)


perrin4869:
Also, btw, maybe it'd be better to change the `displayName` to using parenthesis instead of square brackets?

```js
Translate.displayName = 'Translate(' + getDisplayName(WrappedComponent) + ')';
```

This seems to be more consistent with other HoCs.

jamuhl:
once more...thanks a lot...published: react-i18next@1.6.0

PR #152, opened 2016-07-05
translate HoC displayName change

perrin4869:
This is a minimal change, but I think it's a good idea to follow convention here. For your consideration! Thanks :)


jamuhl:
just changed that...on your last PR's recommendation

Issue #157, opened 2016-07-07
Warning on missing Interpolation prop


aray12:
Is there a way to log a warning if an Interpolation value contains parsed values with no corresponding prop passed. It would be especially awesome if this could tie in with the i18n instance's `debug` setting. Basically I want to easily discover missing props

e.g.
```
{
  "interpolateSample": "you can interpolate {{value}} or {{component}} via interpolate component!"
}
```
```
import React from 'react';
import { translate, Interpolate } from 'react-i18next';

function TranslatableView(props) {
  const { t } = props;

  let interpolateComponent = <strong>a interpolated component</strong>;

  return (
    <div>
      <Interpolate i18nKey='ns:interpolateSample' stupid='"some string"' dumpTypo={interpolateComponent} />
    </div>
  )
}
```

Any thoughts on this?

jamuhl:
on the interpolation component there is the i18n instance added https://github.com/i18next/react-i18next/blob/master/src/interpolate.js#L7

so using the i18n.services.logger.log (warn or error) is easy. If you like you might provide a PR...else i will look into this for future release...just not got the time right now.

PR #160, opened 2016-07-08
Added conditional warning  on unmatched variable during interpolation

aray12:
I mentioned in #157 that it would be nice to receive a warning if no matching prop was provided during interpolation. I noticed that if you use `i18n.t` to try and translate a value that has a variable in it you get the following warning 

```
i18next::interpolator: missed to pass in variable test for interpolating Some value {{test}}
```

I added a basic conditional check that `this.props[match]` actually exists. If it doesn't it calls `this.i18n.services.logger.warn` and essentially mimic the exact same behavior as if you had used that translation with just the standard `i18n.t` function. 

First, I appreciate the work you guys have done. This is definitely the most enjoyable i18n software I have found. Second, I have not committed to OSS much, so forgive me if this is a naive solution. If you have any feedback I will gladly improve on this.

### DIFF
--- a/src/interpolate.js
+++ b/src/interpolate.js
@@ -27,6 +27,7 @@ class Interpolate extends Component {
         child = match;
       } else {
         child = this.props[match];
+        if (!this.props[match]) this.i18n.services.logger.warn('interpolator: missed to pass in variable ' + match + ' for interpolating ' + format);
       }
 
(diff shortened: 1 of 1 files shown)


jamuhl:
looks ok to me...just published react-i18next@1.6.1

PR #194, opened 2016-07-30
Fix bug with undefined store in translate component

newsiberian:
Hello, `this.i18n.store` prop is not always ready/defined to the time when component is loaded. I'm using redux and it occurs with me after React 15.2.1


jamuhl:
did you init i18next before using it - as store gets defined on init.

PR #193, opened 2016-07-28
test: add tests setup + coverage + travis

jmfrancois:
fix #192

Issue #189, opened 2016-07-24
Support raw HTML in translation values


KostyaEsmukov:
Say we have the next translation resource:

    {foo: "<b>Bar!</b>"}

Then `<div><Interpolate i18nKey='foo' /></div>` will result in the next output:
`<div>&lt;b&gt;Bar!&lt;/b&gt;</div>` , while I expect to get this: `<div><b>Bar!</b></div>` , so a raw HTML in translation strings should not be escaped.

Is this an intended behavior, or is a bug? 

I guess replacing [this line](https://github.com/i18next/react-i18next/blob/aa5c10e4b6abbc996f3d428067163900baef16fb/src/interpolate.js#L27) with something like
`child = React.createElement(parent, {dangerouslySetInnerHTML: {__html: match}});` will work this out.

jamuhl:
right changing that line would work it out - but lead to dangerous insert of pure html - depending on interpolating of user inputs would lead to xss vulnerability.

i won't open this door - but you can create a own component doing so - but i would suggest you don't  use html in your translations - better use a markdown component and use markdown...

KostyaEsmukov:
Yeah, I share your concern about allowing use of raw HTML.

However, user input can not be in translation values. So it's only a translator who can put harmful HTML code, but we are supposed to trust our translators and verify their work, aren't we? :)

What do you think about adding a prop like `escapeHTMLinValues` set to true by default? I can submit a PR if you find this appropriate.

jamuhl:
why not just doing it like in the sample: https://github.com/i18next/react-i18next/blob/master/example/app/components/View.js#L19 interpolating a strong component?

KostyaEsmukov:
The `a interpolated component` string is not translated here actually. I need to put some bold parts to a value. Splitting it to multiple keys looks awkward, especially if it is just 3 bold words for a long text.

BTW There's no way to use markdown out of the box, isn't there?

jamuhl:
`<Interpolate i18nKey='common:interpolateSample' component={<strong>{t('key'}</strong>} />`

jamuhl:
no markdown built in...nope

KostyaEsmukov:
Yeah, that's clear.

Another option as for now is
`<div dangerouslySetInnerHTML={{__html: t('foo')}} />`

But they both are weird :-)

jamuhl:
yes...why not using eg. react-remarkable and pass markdown from t function to that?

KostyaEsmukov:
[...]

PR #195, opened 2016-07-31
Added an option of using the <Interpolate /> with a raw HTML

KostyaEsmukov:
Discussion in #189

Usage:

    <Interpolate i18nKey='foo' useDangerouslySetInnerHTML={true} />

TODO:

- [ ] Tests? I'm not sure how to properly test a component, see my comment at #193 
- [ ] Generate dist files

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,13 +155,15 @@ __props__:
 - i18nKey: the key to lookup
 - options: [options](http://i18next.com/docs/options/#t-options) to use for translation (exclude interpolation variables!)
 - parent: optional component to wrap translation into (default 'span')
+- useDangerouslySetInnerHTML: allows use of raw html tags in translation values
+- dangerouslySetInnerHTMLPartElement: optional component to wrap parts of translation values into (default 'span'), used with `useDangerouslySetInnerHTML={true}` only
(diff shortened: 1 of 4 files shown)


jamuhl:
did you init i18next before using it?

KostyaEsmukov:
In my own project? Yes I did, and it works as is. Why do you ask?

In case you need my config, here it is:

    import * as i18n from 'i18next';
    import * as LanguageDetector from 'i18next-browser-languagedetector';

    import { ru } from './ru';

    i18n
      .use(LanguageDetector)
      .init({
        fallbackLng: 'ru',
        whitelist: ['ru'],

        debug: ENV !== 'production',

        interpolation: {
          prefix: '__',
          suffix: '__',
          escapeValue: false  // not needed for react!!
        },
        joinArrays: '\n',
        resources: {
          ru: {
            translation: ru
          }
        }
      });

jamuhl:
sorry landed on wrong PR...did reply on phone while on holiday network...seems mail reply mixed it up...will review PR after holiday.

jamuhl:
published react-i18next@1.7.0. thanks for providing this.

PR #196, opened 2016-08-10
Add translateFuncName option to Translate HOC

rarguelloF:
Hello guys, I've already started using your lib and it's amazing, thanks! 

I've found a little issue using while using it, so that's why I'm creating this PR. 

In the project I'm currently working on, we are extracting the original chains from the source files using `xgettext` and looking for certain function names within the source files, so I found really useful to include an option to specify the name of the property the child components will use for accessing the translation function.

The usage would be the following:

```
export default translate(['namespace'], { wait: true, translateFuncName: 'gettext' })(TranslatedComponent);
```

Cheers and thanks in advance!


jamuhl:
So basically you use `t` function from i18next but prefer to have it named `xgettext`? Or do you not use i18next but some other provider to set the i18n context?

rarguelloF:
Yes, I'm using i18next for translations.

I'm using `xgettext` for extracting the original chains from the source code, and as it's already configured, it's looking for a function named `gettext`. I could change it to look for a function named `t`, but I would have to change other parts of our system which are using `gettext`.

Furthermore, I think a function named `t` is more error prone to dig it in within source code...

jamuhl:
published react-i18next@1.7.0. thanks for providing this.

PR #197, opened 2016-08-19
Support classnames on parents of interpolated components

notmlee:
Hi. There doesn't appear to be a way to customize the `className` prop set on the parent for interpolated components. This is a very naive implementation as I'm not familiar with all of the considerations that need to be made, but I figured it'd be good enough as a starting point for a discussion.


jamuhl:
react-i18next@1.8.0 is published. thanks for contributing.

PR #198, opened 2016-08-19
fix amd build

nassimbenkirane:
fix typo in amd build

by the way, I struggled to make npm build version of react-i18n to work with amd format.
When I use umd version with requirejs (ie. `define('react-i18n', function(reactI18n) {})` ), reactI18n is undefined.

Do you think we could also add `npm run build:amd` to the main build task ?

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
     "build:umd": "rollup -c rollup.config.js --format umd && rollup -c rollup.config.js --format umd --uglify",
     "build:amd": "rollup -c rollup.config.js --format amd && rollup -c rollup.config.js --format amd --uglify",
     "build:iife": "rollup -c rollup.config.js --format iife && rollup -c rollup.config.js --format iife --uglify",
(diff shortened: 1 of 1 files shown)


jamuhl:
yes...does not hurt to add it...ok for me...will do so for next release.

nassimbenkirane:
Great news 👍

jamuhl:
could you test one more thing https://github.com/i18next/react-i18next/blob/master/rollup.config.js#L32 uncommenting that line...as i see this removing that line will change from named export to standard amd export - which should work in your setup?!?

nassimbenkirane:
I won't be able to test it as I'm on m'y way to the holidays airplane, but in order to make it work I indeed had to switch from `import from 'react-i18next' ` to `import from 'reactI18Next'`, which would then be transpiled to the amd define format and consumed by requirejs.

So changing from named export to default amd export should work !

PS : with reactI18Next instead of 'react-18next' it works both with umd and amd builds.

jamuhl:
ok...i will remove named and merge your PR.

PR #203, opened 2016-09-14
Accept i18n context from props or context in translate wrapper

redwoodfavorite:
Useful feature that makes using `translate` HOC easier to use without a provider.  Similar to react-redux: https://github.com/reactjs/react-redux/blob/master/src/components/connect.js#L83


jamuhl:
thank you for contributing...published with v1.10.0

PR #204, opened 2016-09-19
README - Use js instead of json for syntax highlighting

marlonbernardes:
Since JSON does not allow comments, the syntax highlighter was acting a little weird for a block of code contained in the readme file. 

This PR fixes this, by changing the syntax highlighter language from JSON to JS. See images below.

**Before:**
image

**After:**
image


jamuhl:
thanks...actually did not a check when i pushed....

PR #205, opened 2016-09-23
Suppresses required context type warning if passing i18n as a prop

Alex509:
There is a new feature added to pass i18n as a prop (for example using DI) but React emits a warning (Required context type i18n is missing)
References #203


jamuhl:
thank you for the PR. merged and published in react-i18next@1.10.1

PR #221, opened 2017-01-06
Wait for deciding language

Sheile:
When use react-i18next with [i18next-browser-languageDetector](https://github.com/i18next/i18next-browser-languageDetector), My application with following options render text as fallback language while very short period.

### i18next options
```
{ fallbackLng: 'en' }
```

### react-i18next options
```
translate(['common'], { wait: true })
```

### language priority in browser
1. ja
2. en

IMHO, `wait` option should wait for deciding language by languageDetector.

### DIFF
--- a/src/translate.js
+++ b/src/translate.js
@@ -40,8 +40,10 @@ export default function translate(namespaces, options = {}) {
 
         this.mounted = true;
         this.i18n.loadNamespaces(namespaces, () => {
-          if (this.mounted) this.setState({ ready: true });
-          if (wait && this.mounted) bind();
+          this.i18n.changeLanguage(this.i18n.options.lng, () => {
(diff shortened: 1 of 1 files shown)


jamuhl:
did you call i18next.init before passing it to the i18nextProvider? Wondering why `common` namespace is already loaded in `en` for your case but not yet in `ja`. seems more like some configuration issue loadNamespace should call it's callback only if all is loaded.

Can you show me more of your configuration - evtl. your repo?

jamuhl:
did an update:

could you try with:

i18next@4.2.0
react-i18next@2.0.0

should now wait for initialization of i18next instance

Sheile:
Thanks!
Our problem has been resolved at react-i18next@2.0.0.

PR #232, opened 2017-02-17
fix typo in README.md

cwalv:


jamuhl:
thank you for fixing the typo...always glad to get improvement in the texts...as i'm a non native english speaker (but even in my native language spelling is not my strongest talent)

PR #238, opened 2017-03-17
Webpack2 example

nicolechung:
webpack2

Added a working example for webpack 2. Split out the folders so now there a `webpack 1` example and a `webpack 2` example.


jamuhl:
thanks a lot...

PR #239, opened 2017-03-19
typescript tip

nicolechung:
Probably not the best explanation (or any) of why this works...I did a bit of searching but couldn't find an exact explanation of how Typescript `export default` is different from regular ES `export default`


jamuhl:
hm...wonder if we could export more typescript friendly...but honestly got no plan of typescript ;)

but at least this might help users struggling with it - really sorry you lost that much time using our solution.

PR #248, opened 2017-04-10
Fix #247: PropTypes warnings.

### LINKED ISSUE #247: Warning: Accessing PropTypes via the main React package is deprecated. Use the prop-types package from npm instead.
warning.js?8a56:36 Warning: Accessing PropTypes via the main React package is deprecated. Use the prop-types package from npm instead.

The above is the warning that comes after install React `v15.5`. They've moved the `PropTypes` into a separate module called `prop-types`. Please use this.

kadishmal:
Fix #247: PropTypes warnings.


kadishmal:
1. @jamuhl how about incrementing the `react` dependency as well?
2. Will up change the version, or should I do that?

jamuhl:
will check later when out of office and home again - thanks for providing this...really thankful for timesaving - expect it to be published later today.

jamuhl:
1) updating deps is ok for me - normally i do them pre publish - when doing a minor version (not for all patch versions)

Will update the version of react-i18next before publishing

jamuhl:
wonder if we should move react and prop-types to peerDependencies

jamuhl:
just published react-i18next@3.0.0

made it major - so people are aware of needing to have prop-types installed on the main project.

kadishmal:
@jamuhl agree on moving React and and PropTypes into peer deps.

PR #263, opened 2017-06-12
Remove workaround to set ready if there was no initialized signal

jussikinnula:
I have now tested that the problem originally (to need this workaround) was because of misuse of i18next initialization.

The updated code is at:
https://github.com/jussikinnula/react-starter/tree/typescript-i18next

### DIFF
--- a/src/translate.js
+++ b/src/translate.js
@@ -65,10 +65,6 @@ export default function translate(namespaces, options = {}) {
             };
 
             this.i18n.on('initialized', initialized);
-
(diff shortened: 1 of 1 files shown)


jamuhl:
Thank you...wasn't aware i did merge that code part too (wasn't my intend)...published update removing this.

jussikinnula:
I'll keep in mind to make separate PR's for each patch in future, it was my fault too!

jamuhl:
Idk..thought i looked into changes...and there were only the parts for options...never mind...all green now again ;)

PR #278, opened 2017-07-25
Make Trans parent element configurable.

### LINKED ISSUE #276: Trans component should not create a block element
Currently if you use the `Trans` component your text gets wrapped in a `div` element. This is a bit awkward since `div` is a block element, while you are generally looking for an inline element. Wouldn't it be better to use a `span` instead?

wichert:
This fixes #276

### DIFF
--- a/src/trans.js
+++ b/src/trans.js
@@ -101,7 +101,7 @@ export default class Trans extends React.Component {
   }
 
   render() {
-    const { children, count } = this.props;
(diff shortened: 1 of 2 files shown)


jamuhl:
awesome...will check this later. merge and publish an update.

jamuhl:
thank you for the PR. update was published with react-i18next@4.7.0

PR #291, opened 2017-08-28
Expose ES2015 modules as module in package.json

tusbar:
jsnext:main is not supported in webpack (since webpack2).

For reference: https://github.com/webpack/webpack/issues/1979


jamuhl:
Thank you for providing this...forgot to adding it.

published in react-i18next@5.2.0

tusbar:
Thanks @jamuhl!

PR #297, opened 2017-09-12
Pass extra Trans props to parent component

wichert:
This makes it possible to set className or other attributes on the generated parent component. For example you can do use:

```html
<Trans i18nKey="welcome.title" parent="h2" className="highlight">
  Welcome to your new page.
</Trans>
```

to generate:

```html
<h2 class="highlight">Welcome to your new page</h2>
```

### DIFF
--- a/src/trans.js
+++ b/src/trans.js
@@ -117,8 +117,8 @@ export default class Trans extends React.Component {
 
     if (this.i18n.options.react && this.i18n.options.react.exposeNamespace) {
       let ns = typeof this.t.ns === 'string' ? this.t.ns : this.t.ns[0];
-      if (rops.i18nKey && this.i18n.options.nsSeparator && this.i18nKey.indexOf(this.i18n.options.nsSeparator) > -1) {
-        const parts = props.i18nKey.split(this.i18n.options.nsSeparator);
+      if (this.props.i18nKey && this.i18n.options.nsSeparator && this.props.i18nKey.indexOf(this.i18n.options.nsSeparator) > -1) {
+        const parts = this.props.i18nKey.split(this.i18n.options.nsSeparator);
         ns = parts[0];
       }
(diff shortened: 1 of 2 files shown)


src/trans.js
```diff
@@ -109,17 +109,16 @@ export default class Trans extends React.Component {
...
-    const key = this.props.i18nKey || defaultValue;
+    const key = i18nKey || defaultValue;
     const translation = this.t(key, { interpolation: { prefix: '#$?', suffix: '?$#' }, defaultValue, count });
 
-    const additionalProps = {};
     if (this.i18n.options.react && this.i18n.options.react.exposeNamespace) {
       let ns = typeof this.t.ns === 'string' ? this.t.ns : this.t.ns[0];
-      if (this.props.i18nKey && this.i18n.options.nsSeparator && this.props.i18nKey.indexOf(this.i18n.options.nsSeparator) > -1) {
-        const parts = this.props.i18nKey.split(this.i18n.options.nsSeparator);
+      if (rops.i18nKey && this.i18n.options.nsSeparator && this.i18nKey.indexOf(this.i18n.options.nsSeparator) > -1) {
```
jamuhl: could you please fix this to be this.props.i18nKey again

wichert: I can, but I do need to keep the destructuring in place so `i18nKey` does not become part of `additionalProps`, since that will result in a rendering error from React. Since that means i18nKey must be a local variable anyway, I figured that would simplify the spelling here as well.

jamuhl: you also can use that destructed i18nKey. Before it just had a typo `rops.i18nKey` and `props.i18nKey`. Anyway will merge now and publish an update.

wichert: I only noticed those typos later :(. I wonder why they didn't break my testing.

jamuhl:
thank you for the contribution....this addition is published in react-i18next@5.3.0

PR #313, opened 2017-10-02
Update HOC translate method each time i18next resources has changed

rlaranjeira:
Thanks to this change, pure components that needs to render new translation will be able to update.


jamuhl:
see the point - seems the best solution to force a rerender on PureComponents...so will be merged later today and i will publish an update.

Thank you for making me aware of this issue - and providing a fix for it.

jamuhl:
published in react-i18next@6.0.3 - thank you again for providing this.

jamuhl:
sorry react-i18next@6.0.4

PR #315, opened 2017-10-04
Support passing i18n and t as Trans props

dylanseago:
Currently, Trans uses i18n and t props when provided, but also passes them down to the rendered HTML element causing an error.

This PR does the following:
- Adds i18n and t propTypes
- Moves logic that determines which i18n and t to use from constructor to render
- Ensures i18n and t props don't get passed to the HTML element

### DIFF
--- a/src/Trans.js
+++ b/src/Trans.js
@@ -89,32 +89,27 @@ function renderNodes(children, targetString, i18n) {
 
 export default class Trans extends React.PureComponent {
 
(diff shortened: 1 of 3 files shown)


jamuhl:
will have a look and merge tomorrow...thanks for providing this PR

jamuhl:
Thanks again...published in  react-i18next@6.0.5

PR #328, opened 2017-10-19
Change prop-types from a peer dependency to a normal dependency

TLadd:
Recommended by the prop-types package: https://github.com/facebook/prop-types#how-to-depend-on-this-package

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,8 @@
   },
   "dependencies": {
     "hoist-non-react-statics": "2.3.1",
-    "html-parse-stringify2": "2.0.1"
+    "html-parse-stringify2": "2.0.1",
+    "prop-types": "^15.6.0"
   },
   "devDependencies": {
     "babel-cli": "6.26.0",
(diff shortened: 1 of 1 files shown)


jamuhl:
@TLadd do you know why this is recommended? Looks opposite to what we used to do...would be great to know the reasons - not just having a recommendation...

TLadd:
@jamuhl Here's a few issues where this was discussed: 
https://github.com/facebook/prop-types/issues/44
https://github.com/facebook/prop-types/issues/4

With react and react-dom, the app needs to be in control of what versions are being used. Having multiple versions loaded will even break the app. With prop-types, however, there is no such restriction. As a user of the library, I don't care what version of prop-types is being used.

jamuhl:
thank you for taking the time to pass me the additional information. Merged now. Would you need a publish of this on npm - or ok if it comes with next release?

TLadd:
Waiting for the next release is fine with me. Thank you!

PR #337, opened 2017-11-02
Add basic test for <I18n /> component

jonathaningram:
The component appeared to have some test coverage but was not directly
tested. This suite can be expanded to include more tests.

Aside: I've been trying to get my project's tests working for components that use `<I18n />` and have not had any luck so far. When I checked the `react-i18next` repo to see if it had any tests for this, I couldn't find any so decided to write one. I might provide another PR with some more advanced tests as I start to debug my own project.

Thanks.

### DIFF
--- a/test/i18n.render.spec.js
+++ b/test/i18n.render.spec.js
@@ -5,7 +5,7 @@ import I18n from '../src/I18n';
 
 const context = { i18n };
 
-describe('I18n simple', () => {
+describe('I18n', () => {
   it('should render correct content', () => {
     const wrapper = mount(
       <I18n ns="translation">{t => <span>{t('key1')}</span>}</I18n>,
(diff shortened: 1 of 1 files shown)


jonathaningram:
OK I added a failing test case. @jamuhl could you please take a look at the second test and tell me if it looks like a bug or if I'm doing something wrong? The bug stack can be seen here: https://travis-ci.org/i18next/react-i18next/builds/296088967#L1835 I will wait for your input, but from my debugging I see that `context` (and thus `context.i18n`) is not actually available in the constructor for some reason.

Also I didn't realise that @andrewbranch had created https://github.com/i18next/react-i18next/pull/336. Depending on what goes in first, I'm happy to rebase this. FWIW, it looks like Andrew's tests are more whitebox whereas I'm testing usage when actually rendering the component.

jamuhl:
will need to look deeper into this...think i will add your tests after merging andrewbranch his PR

PR #339, opened 2017-11-02
Remove i18n prop from <I18n /> tests

jonathaningram:
The prop should not need to be passed to get this work.

See https://react.i18next.com/components/i18n-render-prop.html. You don't have to pass the prop in an app to get the render prop to work, so I'm not sure why you have to pass it in the test.

Thus, passing the prop in the test is actually a false positive.

@jamuhl correct me if I'm wrong, but I don't think your commit actually fixes the problem, it just silences the test: https://github.com/i18next/react-i18next/commit/0cd849d7070335adbaf508db34e945365ddf6448#diff-798d3ecb9f34ff4cd22a54a7e848c0f5R22

### DIFF
--- a/test/i18n.render.spec.js
+++ b/test/i18n.render.spec.js
@@ -8,7 +8,7 @@ const context = { i18n };
 describe.only('I18n', () => {
   it('should render correct content', () => {
     const wrapper = mount(
-      <I18n i18n={i18n} ns="translation">{t => <span>{t('key1')}</span>}</I18n>,
+      <I18n ns="translation">{t => <span>{t('key1')}</span>}</I18n>,
       { context }
     );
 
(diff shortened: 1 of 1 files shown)


jamuhl:
problem is only from this test i get context {} while every other test seems to pass context proper...not sure it's a enzyme thing....at least currently i get it not to work with no i18n in props...but seems a test issue not a code issue.

jonathaningram:
Yeah it could be an enzyme thing, but I feel that we should aim to make the test represent how the component should actually be used. Especially because some users will look at the tests as _the_ documentation. I can try a few things myself to see if it's possibly an enzyme thing too. E.g. we can try react shallow renderer.

jamuhl:
agree...using is one of the options - agreeing on passing it by props is the "last used option". The other options is by using this https://github.com/i18next/react-i18next/blob/master/example/react_renderProps/src/i18n.js#L9 using react-i18next internal state which enables using react-i18next without the i18nextProvider (which i personally prefer - but both options are valid)

But fully agree if we get it to work - that would be awesome - as it leaves some strange feelings that it obviously does not work in tests...

jamuhl:
cosing this...so much changed on the way to v8.0.0

PR #353, opened 2017-12-02
Use t from I18n if not present in context

### LINKED ISSUE #352: t not set if Trans is used only with I18nextProvider
If the Trans component is used outside of a component wrapped with the I18nHoc but in an application using the I18nextProvider, then `t` is undefined (resulting in a render error). 

```jsx
<I18nextProvider i18n={i18n}>
  <Trans key="Test" />
</I18nextProvider>
```

The issue is, that the I18nextProvider only sets the i18n object into the context but `Trans` expects `t` and `i18n` to be present. I'm willing to fix this issue in a PR but first wanted to ask what the preferred way to solve the issue is:

1. Also set `t` in I18nextProvider. 
2. Mark `t` as not required in `Trans` and use `i18n.t` if  `t` is not explicitly defined

Which solution would you prefer?

MichaReiser:
Use `i18n.t` in the Trans component if `t` is not set with props or present in the context.
This allows to use Trans with the I18nextProvider.

Fixes #352

### DIFF
--- a/src/Trans.js
+++ b/src/Trans.js
@@ -96,7 +96,8 @@ export default class Trans extends React.Component {
 
   render() {
     const contextAndProps = { i18n: this.context.i18n, t: this.context.t, ...this.props };
-    const { children, count, parent, i18nKey, i18n, t, ...additionalProps } = contextAndProps;
+    const { children, count, parent, i18nKey, i18n, t: tFromContextAndProps, ...additionalProps } = contextAndProps;
+    const t = tFromContextAndProps || i18n.t.bind(i18n);
 
     const reactI18nextOptions = (i18n.options && i18n.options.react) || {};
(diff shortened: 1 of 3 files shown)


src/Trans.js
```diff
@@ -96,7 +96,8 @@ export default class Trans extends React.Component {
 
   render() {
     const contextAndProps = { i18n: this.context.i18n, t: this.context.t, ...this.props };
-    const { children, count, parent, i18nKey, i18n, t, ...additionalProps } = contextAndProps;
+    const { children, count, parent, i18nKey, i18n, t: tFromContextAndProps, ...additionalProps } = contextAndProps;
```
jamuhl: what is tFromContextAndProps? that seems not to be defined?

MichaReiser:
Thx!

jamuhl:
published in react-i18next@7.1.0

thank you for contributing

PR #364, opened 2017-12-12
Render empty string on empty string as Trans child.

### LINKED ISSUE #363: Crash on empty text nodes
Happens for this:

`<Trans>{""}</Trans>`

This is pretty annoying if you have an optional message somewhere:

```ts
const message = success ? "Succeeded." : null;
return <Trans>{message || ""}</Trans>;
```

```
Trans.js:85 Uncaught TypeError: Cannot read property 'content' of undefined
    at Trans.js:85
    at Array.reduce (<anonymous>)
    at mapAST (Trans.js:73)
    at renderNodes (Trans.js:100)
    at Trans.render (Trans.js:143)
```

phiresky:
Closes #363

### DIFF
--- a/src/Trans.js
+++ b/src/Trans.js
@@ -47,6 +47,7 @@ function nodesToString(mem, children, index) {
 }
 
 function renderNodes(children, targetString, i18n) {
+  if (targetString === "") return [];
 
   // parse ast from string with additional wrapper tag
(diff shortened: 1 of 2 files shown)


jamuhl:
published with react-i18next@7.2.0

Issue #369, opened 2017-12-14
Allow first argument in translate HOC to be a function


EmmaSimon:
Would it be possible for `translate` to take a function that returns the namespace(s)? The function would be passed the props of the component. It currently doesn't seem possible to use `translate` to conditionally load namespaces from the backend based on the wrapped component's props.

jamuhl:
does that function have to be async? a simple synchronous function would be not a big deal...but an async function while possible is rather tricky to get right...

jamuhl:
so

`translate(() => { return ['ns1', 'ns2'] })(mycomponent)`

or:

`translate((callback) => { callback(['ns1', 'ns2']) })(mycomponent)`

EmmaSimon:
Oh, no I think synchronous is fine. But I'd think this would only really be useful if it was provided the props. So you can do something like,
```
translate((props) => {
    if (props.x) {
        return ['ns1'];
    }
    return [`ns-${props.y}`];
})(mycomponent)
```

jamuhl:
that's doable: https://github.com/i18next/react-i18next/blob/master/src/translate.js#L21

if you like feel free to make a pull request...else i will add it sometime between xmas and new year.

PR #372, opened 2017-12-15
Allow translate HOC to take a function

EmmaSimon:
For #369. It was slightly more complicated than just adding `if (typeof namespaces === 'function')             namespaces = namespaces(props);`, because when multiple of the same component were rendered, only the first one would run the function, then the closure in the HOC would have that result set to `namespaces` for every other.

I'm not certain if `Translate.namespaces` was being used elsewhere, but now that value can also be a function. If that'll cause issues, let me know and I'll try to implement it another way.

### DIFF
--- a/src/translate.js
+++ b/src/translate.js
@@ -9,7 +9,7 @@ function getDisplayName(component) {
   return component.displayName || component.name || 'Component';
 }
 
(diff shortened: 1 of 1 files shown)


jamuhl:
published with react-i18next@7.3.0

thank you for contributing

PR #411, opened 2018-03-21
Fix reference to props in constructor

mjfaga:
In <=IE10, `this.props` is null, so when attempting to set the namespaces,
the reference to `this.props.ns` raises an exception. By updating the
reference to use the version passed into the constructor, we are
consistent with other usages of props on the constructor, and I've
confirmed this patch works in <=IE10 as well.


jamuhl:
thank you...published in react-i18next@7.5.1

PR #416, opened 2018-04-06
Nested keys

AndruBot:
Nested keys

### DIFF
--- a/src/Interpolate.js
+++ b/src/Interpolate.js
@@ -1,5 +1,6 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
+import get from 'lodash/get';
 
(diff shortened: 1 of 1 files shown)


jamuhl:
Just had a fast look....if i get it right it allows for nested objects in the Interpolate component plus adds support for unescaping those

Beside Interpolate will be deprecated i guess as Trans component does same better: https://react.i18next.com/components/trans-component.html

jamuhl:
Would accept this PR if removing dependency on lodash/get

jamuhl:
Interpolate is marked deprecated in v8.0.0 and will be removed in v9.0.0 -> so closing this as it does not make a lot of sense anymore

PR #435, opened 2018-05-15
Replace @kadira/storybook to @storybook/react

9renpoto:
`@kadira/storybook` renamed `@storybook/react`


jamuhl:
did not test - but i guess sample still works ... thank you for the update

PR #462, opened 2018-06-28
In Trans.js, set default value to undefined as last possible option

dominicboutin:
Using Trans component, if node is left empty, default value is set to a empty string. I18next will use that default value instead of its fallback mechanism. By setting the default value to undefined will let i18next define what fallback to use.

### DIFF
--- a/src/Trans.js
+++ b/src/Trans.js
@@ -105,7 +105,7 @@ export default class Trans extends React.Component {
     const reactI18nextOptions = (i18n.options && i18n.options.react) || {};
     const useAsParent = parent !== undefined ? parent : reactI18nextOptions.defaultTransParent;
 
-    const defaultValue = defaults || nodesToString('', children, 0);
+    const defaultValue = defaults || nodesToString('', children, 0) || undefined;
     const hashTransKey = reactI18nextOptions.hashTransKey;
     const key = i18nKey || (hashTransKey ? hashTransKey(defaultValue) : defaultValue);
     const interpolationOverride = values ? {} : { interpolation: { prefix: '#$?', suffix: '?$#' } };


jamuhl:
isn't defaultValue not just used in case of the fallback not finding something: https://github.com/i18next/i18next/blob/master/src/Translator.js#L122

dominicboutin:
Exactly, since defaultValue is set to a empty string and res is undefined, 
https://github.com/i18next/i18next/blob/master/src/Translator.js#L126
used isn't set to true
res isn't set to key

plus that skip the parseMissingKeyHandler function
https://github.com/i18next/i18next/blob/master/src/Translator.js#L177

jamuhl:
beside how are you using the Trans component to not have a meaningful defaultValue...do you use it everywhere to just render plain strings not having a component interpolated? if so that could be done by just using the t function from the hoc, render prop

dominicboutin:
We are using TypeScript, its easier to use Trans in our components than set the prop type to include t. 
`<Trans i18nKey="somekey" />`

Setting returnEmptyString to false fixes our issues but I feel like the default Trans value should undefined. 
Adding a reactI18nextOptions.defaultTransDefaultValueFallback could be a solution or wait for the next major version.

On our side the issue is fixed with that workaround, this PR can be closed.

jamuhl:
will keep this open for now - will add the `defaultTransDefaultValueFallback` asap. So you got that option.

jamuhl:
closing this for now...added an issue to keep that for future development: https://github.com/i18next/react-i18next/issues/577

sorry for not yet having time to do so...a PR would be welcome

jamuhl:
published in react-i18next@9.0.0

allow defining `transEmptyNodeValue` in i18next.options.react to set a default fallback value

PR #474, opened 2018-07-12
Don't require `i18n` in context

### LINKED ISSUE #473: Don't warn about `i18n` not being present in context
When passing `i18n` as a prop to the `Trans` component, it still warns about `i18n` not being present in context. This is annoying and could be fixed by either making it not required, or by using a custom prop type function to check if it's present `props`.

kevva:
Since it can be passed in as a prop, we shouldn't require it to be present in context.

Fixes #473.

### DIFF
--- a/src/Trans.js
+++ b/src/Trans.js
@@ -143,6 +143,6 @@ Trans.propTypes = {
 // };
 
 Trans.contextTypes = {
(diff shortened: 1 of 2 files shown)


jamuhl:
published in react-i18next@7.8.1

PR #476, opened 2018-07-15
Don't polyfill Object.entries globally

goto-bus-stop:
Patching `Object.entries` affects other code beyond this library, which
ideally should not happen. With this patch the polyfill is still used
when necessary but the global environment is not unexpectedly changed by
including react-i18next :)


jamuhl:
will be merged and published with next version

PR #478, opened 2018-07-17
Support defaultNS from I18nextProvider props

iancw:
This allows the default namespace to be specified as a prop of
`I18nextProvider`. Previously namespace came from i18n instance option,
and then from arguments to the `translate` HOC. This allows the
`translate` HOC to be invoked without an explicit namespace, and the
parent `I18nextProvider` can define an alternate default namespace. This
allows sharing `translate`d components without hard coding a defaultNS
at build time and without cloning multiple i18n instances to set
alternate namespaces.

e.g.

```
const Translated = translate()(Elem);

<I18nextProvider defaultNS="namespaceOne">
    <Translated />
</I18nextProvider>

<I18nextProvider defaultNS="namespaceTwo">
    <Translated />
</I18nextProvider>
```

### DIFF
--- a/src/I18nextProvider.js
+++ b/src/I18nextProvider.js
@@ -45,4 +45,8 @@ I18nextProvider.childContextTypes = {
   defaultNS: PropTypes.string
 };
 
+I18nextProvider.defaultProps = {
+  defaultNS: undefined
+};
+
 export default I18nextProvider;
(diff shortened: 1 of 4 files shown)


jamuhl:
Awesome...published in react-i18next@7.9.0

jamuhl:
Good addition - might be this is one of the last versions before breaking to use the new context API.

PR #482, opened 2018-07-25
Skipped custom html tags

hrnik:
## Description

Fixed the bug, with empty html tags in source for Trans component. 
Translators can (yes, they can) add some html tag into translations.

For example we had source 
```js
{
 "key" : "Some text <1>other text<1>"
}
```

But translations manager decide to add some tags in translations:
`"Some text <br /> <1>other text<1>"`

After that react-i18next throw the error

```
    TypeError: Cannot read property 'content' of undefined

      at src/Trans.js:78:86
          at Array.reduce (<anonymous>)
```

### DIFF
--- a/src/Trans.js
+++ b/src/Trans.js
@@ -75,8 +75,11 @@ function renderNodes(children, targetString, i18n) {
             inner
           ));
         } else if (typeof child === 'object' && !isElement) {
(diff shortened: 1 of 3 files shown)


jamuhl:
thank you...published in react-i18next@7.9.1

Issue #485, opened 2018-07-31
DRY issue with the razzle-ssr example


a-barbieri:
Hi I have a DRY issue with the **razzle-ssr example**. 

I can probably solve it myself but I'd rather ask before doing something wrong.

1. Is there any reason for which [i18n options](https://github.com/i18next/react-i18next/blob/master/example/razzle-ssr/src/i18n.js#L5-L29) aren't implemented [on the server side rendering](https://github.com/i18next/react-i18next/blob/master/example/razzle-ssr/src/server.js#L21-L25)?
1. Is there a reason for which `wait` isn't set globally among [i18n options](https://github.com/i18next/react-i18next/blob/master/example/razzle-ssr/src/i18n.js#L5-L29), but spread around the application ([here](https://github.com/i18next/react-i18next/blob/master/example/razzle-ssr/src/i18n.js#L32) and [here](https://github.com/i18next/react-i18next/blob/master/example/razzle-ssr/src/Home.js#L33))

Happy to make a PR if needed

jamuhl:
1) init is here and than here again for server specific stuff -> doing a second init will use existing options and just overwrite the new options

2) might be cause that global option did not yet exist when creating the sample...or just some copy over...could be changed to global if you prefer - PR welcome

a-barbieri:
> doing a second init will use existing options and just overwrite the new options
Good to know.

I've made a PR to update the example and DRY `wait` option

jamuhl:
merged...thank you

PR #486, opened 2018-08-01
DRY i18next/react-i18next#485

a-barbieri:
i18next/react-i18next#485

### Updates
1. Update dependencies
1. Make `wait` option gloabl
1. FIX Razzle server config which was missing locales files


jamuhl:
awesome .... thank you

Issue #504, opened 2018-09-03
I18NextProvider.reportNS function does not report default namesapce


tutok:
It looks like namespaces reported by `reportNS` is `undefined` when no namespace is specified and fallback rule is executed. 

For example, if I set `{ defaultNS: 'global' }` option in i18next and use translate HOC `translate()(Home);` then component `Home` will be translated using `global` namespace. But `reportNS` returns `undefined`

@olivierpascal can you look on this?

olivierpascal:
This is definitely a bug. I'll take care of it.

olivierpascal:
@tutok are you using this feature for SSR? Just wondering.

tutok:
@olivierpascal yes, I use it with SSR

jamuhl:
guess this one could be closed...PR was merged

PR #506, opened 2018-09-03
related to issue #504: also report default Namespace with translate()

olivierpascal:
I also chose to report `undefined` when no namespace is used with `translate()` and no default namespace is defined in options.


jamuhl:
thanks again - was published in react-i18next@7.11.1

Issue #508, opened 2018-09-05
PrevProps the same as current props


lomboboo:
Hi and thanks for awesome library.

I use react-i18next in my project and I've got an issue when I want to access `prevProps` within `ComponentDidUpdate` for example. In this case `language` field in the `i18n` object is the same as `this.props`. You can see it in this [demo](https://codesandbox.io/s/03877y0960) .

I took a look in the [core](https://github.com/i18next/react-i18next/blob/master/src/translate.js#L20) and found that inside `translate` HOC you mutate `i18n` object. 

Is there any reason why it was implemented this way? I'd really like to have `prevProps` has correct values

jamuhl:
it is the same object (singleton) and not an immutable...it just try to look it up in the different possible locations (context, props, ...)

what exactly is your issue - why you need that previous language?

lomboboo:
Thanks for quick response.

I have a case where I need to do some async calls when language changed. So I'd have condition in the `ComponentDidMount` something like `this.props.i18n.language !== prevProps.i18n.language` then do some logic.

Although i18next has `languageChanged` event, it's inconvenient to use in the react app. It'd be much simpler if prevProps worked "react like".

jamuhl:
hm...then i would suggest we pass lng from translate to component as a prop...the i18n instance is based on i18next and can't be really changed therefor

so adding it to the I18n render props which is used in translate hoc and from there add it to components props?!?

lomboboo:
I can work on PR, but I'm not sure what are you referring to. You're talking about pass as prop [here](https://github.com/i18next/react-i18next/blob/master/src/translate.js#L74)?

jamuhl:
yes and also here: https://github.com/i18next/react-i18next/blob/master/src/I18n.js#L124 

basically passing it from that I18n to the line you showed above...

feel free to make a PR

PR #511, opened 2018-09-07
related to issue #508: react_i18n object property

lomboboo:
This PR relates to #508 . It gives ability to compare `language` prop with the prevProps or nextProps in the react lifecycle hooks.

I thought it'd be better to pass `react_i18n` object property (which contains language) instead of raw `language` because it can be used in the future for other needs.


jamuhl:
published in react-i18next@7.12.0

with one change - i prefer having a prop `lng` passed down instead of a react_i18n prop (we might add this later for less important stuff - if we ever need todo so).

PR #530, opened 2018-10-02
Add render prop example to Storybook integration

lonyele:
I had little problems while integrating with storybook. Mostly It is trivial, but for those whom are same as me can find it helpful. Just feel free to correct me if something is wrong

### DIFF
--- a/example/storybook/README.md
+++ b/example/storybook/README.md
@@ -16,3 +16,5 @@ This project was bootstrapped with [Create React App](https://github.com/faceboo
 ## information
 
 In the sample we show both a component being decorated by translate hoc using a [storybook decorator](https://storybook.js.org/basics/writing-stories/#using-decorators) to wrap the `I18nextProvider` around the decorated component and a pure component just expecting a t function which we pass in with a mock function.
+
+If you prefer a render prop style, you need to import and use reactI18nextModule on i18n.js and import '../src/i18n' on the storybook file.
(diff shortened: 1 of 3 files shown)


jamuhl:
should not be needed...had not yet time to upgrade to v8 of react-i18next -> will need to update that and check again if things work -> but yes...removing the I18nextProvider and using the i18nextReactModule would be nicer

jamuhl:
not needed refers to the use(reactI18nextModule) as the decorator I18nextProvider already does the trick...but your solution will work

using the I18n render prop is just needed as SayWelcome itself is not decorated with the translate hoc (just a pure component)

lonyele:
> not needed refers to the use(reactI18nextModule) as the decorator I18nextProvider already does the trick...but your solution will work
> 
> using the I18n render prop is just needed as SayWelcome itself is not decorated with the translate hoc (just a pure component)

Oh, I put 3 examples in one so it kind of look messy... It would have been clear If I divided all 3 into files.
1. HOC style and not decorated example are same as yours
2. However think of clean repository with only using render prop, `i18n.use(reactI18nextModule)` and `import '../src/i18n'` are definitely needed.(because in this scenario, I didnt' put  `I18nextProvider ` or wrap with `tranlate()`) These are the point that I wanted to address. 

Yeah, just feel free to accept this or not. I just tried to wrap my head around.

jamuhl:
redone the samples for v8...https://github.com/i18next/react-i18next/tree/master/example/storybook/stories

would love to hear your opinion and possible options for improvement...

Issue #535, opened 2018-10-04
NextJs example flickers on page load


isaachinman:
I seem to have great timing - I'm just now looking for a localisation solution for NextJs and see you've updated the examples here within the past 24hrs.

However, knowing personally how hard localisation can be with SSR, I downloaded the `react-i18next/example/nextjs` dir, and did `yarn && yarn build && yarn start`.

If you then open `localhost:3000`, you'll see the page loads, then flashes white, then loads again.

Not sure exactly what's happening here, but the example can't be taken seriously in its present state.

**Update:** If I remove `NamespacesConsumer` from `_app.js` and render `<Component {...pageProps} />` directly, the flicker stops. So the problem lies somewhere therein.

**Further update:** If I swap the order of your `i18n` prop pipe on `NamespacesConsumer`, the flicker stops:

> i18n={i18n || (pageProps && pageProps.i18n)}

However, the flicker seems to stop because the initial render just isn't happening. In other words, the page simply remains white until the second render happens. This is not really a fix as it seems to negate SSR. I've noticed that `pageProps.i18n` is `null` on my clientside render, whereas it's a full object on my serverside render.

It would be great to get a response on this as soon as the maintainers are able.

jamuhl:
retry now...just fixed the problem in _app.js: https://github.com/i18next/react-i18next/commit/c456a57a3504b2c0a9a5c1cf7eeebb21b074550e

isaachinman:
@jamuhl Wow, thanks for a lightening fast response.

That does indeed seem to fix it. I had a feeling it was related to this `wait` prop. What are the implications of disabling `wait`?

Also just a small thing, it might be more helpful to specifically pass the necessary props into `NamespacesConsumer` rather than anonymously spreading them in, as this is an example after all.

isaachinman:
Also if you don't mind, it would be useful to understand how one would dynamically change the locale clientside in this setup.

jamuhl:
yes...like you said i18n and ssr is hard - so it is to create a sample that is still readable...will try to improve (but honestly currently i need a step back - worked to hard to get v8 of react-i18next out and adjusting all samples -> which was not so hard because of v8 of react-i18next but all other frameworks like next.js evolving too)

[...]

Issue #499, opened 2018-08-23
SSR with exact namespaces loaded


olivierpascal:
I would like to propose a simple way to pass down only necessary namespaces to client.

```javascript
const namespaces = new Set();

const App = () => (
  <I18nextProvider i18n={I18n} reportNS={ns => namespaces.add(ns)}>
    {bodyComponent}
  </I18nextProvider>
);

renderToString(<App />);

// Now we can use `namespaces` to construct a initialI18nStore object
// and pass it down to client through HTML. Example:

const initialLanguage = 'en';
const initialI18nStore = { [initialLanguage]: {} };

// i18nInstance is a i18next instance with all languages and all namespaces preloaded
Array.from(namespaces).forEach(ns => {
  initialI18nStore[initialLanguage][ns] =
    i18nInstance.services.resourceStore.data[initialLanguage][ns];
});

// Then inject in HTML:
// window.initialI18nStore = JSON.parse('${JSON.stringify(initialI18nStore)}');
```

Would you accept a PR?

jamuhl:
Seems to be reasonable - for sure would accept a PR.

PR #538, opened 2018-10-06
Add language subpaths to NextJs example

isaachinman:


isaachinman:
@jamuhl Here's a possible solution to what we were discussing in #535.

This is an opinionated approach, but then again any approach would have to be. In general, here are the things we needed to do:

1. Allow NextJs to render identically for `/page1` and `/de/page1`
2. Allow i18n to understand (and set) the locale difference between `/page1` and `/de/page1` during SSR
3. Allow client to watch for all route changes, and make sure the incoming path contains the correct (if any) language subpath
4. Allow client to watch for all locale changes, and make sure the incoming path contains the correct (if any) language subpath

Generally speaking, the first two are server-oriented and the relevant changes are in `server.js`.

The second two are client-oriented and the relevant changes are in `pages/_app.js`, as this is the "entry point" for our client.

I've also added a config file, where you can define your default language, and all other languages.

The end result:

    /page1 (English)
    /de/page1 (German)
    /de (Redirects to /de/ to make parsing more dependable)

Changing the route via the NextJs router directly, or changing the locale via the i18n instance directly will both result in validation of the incoming route against current or incoming locale. It's important to note that I can just write a `<Link href="/page3">` or `Router.push("/page3")`, without having to worry about current language, and our middleware will set the correct localised path. Another nice added benefit is that locale changes are also persisted across browser history. If you change locales, you can hit the back button on your browser and go back to the previous locale.

Side note: I'm not really sure what is happening with `pages/page3.js`, but the h1 title breaks. I think this was something you already noticed previously.

Hopefully all of this makes sense. Let me know what you think. There are probably some improvements to be made here, but most likely I will be building a production application taking this approach shortly.

jamuhl:
🙌 you get the "i18next hero of the month" badge - this is awesome

example/nextjs/server.js
```diff
@@ -39,8 +61,34 @@ i18n
         // missing keys
         server.post('/locales/add/:lng/:ns', i18nextMiddleware.missingKeyHandler(i18n));
 
+        // Add language subpaths via i18n middleware
+        i18nextMiddleware.addRoute(
```
jamuhl: is this needed? does not the * route below with the next.js handle work alone

isaachinman: Yes, totally correct! Forgot about that block, I should have removed it.

isaachinman:
Glad to help. Not sure if you think it's smart to merge this, because it does make the example a bit more complicated... If anyone has any questions about how it works they can contact me directly.

isaachinman:
@jamuhl I've come across a strange SSR bug, where namespaces don't seem to load correctly, and I was wondering if you can help me debug it.

[I've created a branch on my fork here](https://github.com/isaachinman/react-i18next/tree/ssr-rendering-broken) that demonstrates the problem as simply as possible.

Commenting and uncommenting out [this line](https://github.com/isaachinman/react-i18next/blob/ssr-rendering-broken/example/nextjs/pages/index.js#L10) seems to break the loading of the `common` namespace for the entire app, despite `_app.js` having its own `NamespacesConsumer` which is supposed to be loading in the `common` namespace.

Moreover, even if I wrap `Navbar` in its own `withI18next(['common'])` HOC instead of taking `t` from props, our SSR is still broken.

Any idea what's going on here? My initial impression was that our app could contain as many `NamespacesConsumer` as we desired, and we could connect components arbitrarily to different namespaces via the `withI18Next` util from any point in the component tree.

I would appreciate any insight you can offer, I have a feeling my lack of understanding comes from lack of experience in the i18next ecosystem in general.

isaachinman:
Tracked it down to [this function](https://github.com/isaachinman/react-i18next/blob/master/example/nextjs/i18n.js#L42).

Somehow, despite the React tree being rendered containing more than one `NamespacesConsumer`, calling for more than one namespace, the second argument of `i18n.getInitialProps` is a single namespace.

Any idea why? Is it possible this is a bug within `react-i18next`?

jamuhl:
Will have to look deeper into this on Tuesday...won't have time tomorrow...my little girl has her birthday.

The problem - i guess is in the function you mention - the basic sample just uses an approach which expects the page level withI18n hoc to contain all namespaces.

There is a function you can pass down `reportNS` you might incorporate that into the withI18n hoc -> the functionality is like: https://github.com/i18next/react-i18next/issues/499

tests for it: https://github.com/i18next/react-i18next/blob/master/test/withNamespaces.spec.js#L72

isaachinman:
Ah yeah, seems like a simple misunderstanding. I at first though `i18n.getInitialProps` was a built-in that tied into `NamespacesConsumer` via `react-i18next`.

So, the problem generally speaking is: we need a way to traverse a React component tree, and determine what locales are required to render that tree correctly.

The suggestion from @olivierpascal in #499 looks interesting, but the performance of `renderToString` is famously bad. Did @olivierpascal mean that `react-i18next` would do a `renderToString` _just_ to determine the namespaces? If `renderToString` is the only approach, it seems like that belongs in userland. No one should be calling `renderToString` twice per request.

Perhaps it is a strange idea, but what about using something like `react-test-renderer`, and then recursively checking the JSON structure for namespace props?

I imagine that this will be necessary pretty quickly, when using the approach in this PR/example to build out an application of decent size. The goal is to be able to declare namespace dependencies via a decorator on each individual component, and have reconciliation/SSR happen in the most efficient way possible.

olivierpascal:
@isaachinman the code you saw in #499 is supposed to be a server-side code for SSR purpose. The `renderToString` is not specifically used to report namespaces. This is the `renderToString` you already use if you are doing SSR.

isaachinman:
So, does the top-level `reportNS` give _all_ namespaces within a tree already?

olivierpascal:
Yep

isaachinman:
Oh, that's great news. Thanks, that should be all we need here.

isaachinman:
@jamuhl Let me know if you had an idea about how to use `I18nextProvider` in our current NextJs situation, it's not clear to me.

jamuhl:
The provider is not needed in first place -> the withI18n takes the role of the provider passing down the i18n instance: https://github.com/i18next/react-i18next/blob/master/example/nextjs/lib/withI18next.js#L12

jamuhl:
@isaachinman did you had luck with using the reportNS function?

isaachinman:
@jamuhl Unfortunately I haven't had a chance yet - very busy week. I think it's worthwhile to get it working for this example though, or lots of people will hit the same confusion. Will update here when I get a chance to look at it.

Sergey2095:
The problem with this implementation is that the page reloads when it has to work on the client side.

isaachinman:
@Sergey2095 That is correct, when switching from default to non-default locale there will be a document request. I believe this is due to the way NextJs intercepts Express requests.

I was aware of this whilst building out the functionality, but do not currently know a way around it. I'd be very appreciative of any advice anyone might have.

Although it is an annoyance, my general outlook is that locale switching is not some "high performance", frequent activity that should retain SPA strictly. In reality this PR is mostly about routing and SEO concerns.

Again, happy to accept any advice. It's most likely a trivial thing about how the `Router.events.on('routeChangeStart')` function block is working.

isaachinman:
@jamuhl I have some time to look at this `reportNS` solution now. What did you have in mind? How are we going to compile the argument returned from `reportNS` across multiple HOCs and pass that into `getInitialProps`?

jamuhl:
merged this and made it with a config setting to opt-out: https://github.com/i18next/react-i18next/blob/master/example/nextjs/config.js#L10

Issue #554, opened 2018-10-18
NextJS issue after changing locale


revskill10:
In latest NextJS example, after changing locale, and refresh browser, there're two issues:
- The locale keeps using old locale before switching to new locale. 
- There's a mismatch between server content (using old locale) and the client content (using new locale)

Note: If I set `enableSubpath` to `false`, the second issue disappear, but the first one keeps using old locale before switching to new locale.

revskill10:
@jamuhl  Ah, the issue is invalid because it requires you to use cookie caching if you want to reuse lang when switching.

jamuhl:
hm, what do you think...should we enable that in the sample? i guess others will not easily find that cookies need to be enabled to get that back to serverside?

revskill10:
@jamuhl I found the solution in one of closed issues about caching with cookies. 
But for newscomers who wants everything just works out of the box, i think we should enable it.
Do you want me to make a PR ?

jamuhl:
agree...a PR would be awesome...would save me some time....👍

PR #555, opened 2018-10-19
Use localStorage and cookie for caching locale

revskill10:
https://github.com/i18next/react-i18next/issues/554

### DIFF
--- a/example/nextjs/pages/_app.js
+++ b/example/nextjs/pages/_app.js
@@ -5,6 +5,7 @@ import { NamespacesConsumer } from 'react-i18next';
 import i18n from '../i18n';
 import languagePathCorrection from '../lib/languagePathCorrection';
 import { translation } from '../config';
+import 'cookie' from 'cookie'
 
 const { enableSubpaths } = translation;
 
@@ -48,7 +49,9 @@ export default class MyApp extends App {
               <button
                 type="button"
                 onClick={() => {
-                  i18n.changeLanguage(i18n.languages[0] === 'en' ? 'de' : 'en');
+                  const lng = i18n.languages[0] === 'en' ? 'de' : 'en'
+                  document.cookie = cookie.serialize('i18next', lng)
+                  i18n.changeLanguage(lng)
                 }}
               >
                 Change locale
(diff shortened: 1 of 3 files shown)


example/nextjs/pages/_app.js
```diff
@@ -48,7 +49,9 @@ export default class MyApp extends App {
               <button
                 type="button"
                 onClick={() => {
-                  i18n.changeLanguage(i18n.languages[0] === 'en' ? 'de' : 'en');
+                  const lng = i18n.languages[0] === 'en' ? 'de' : 'en'
+                  document.cookie = cookie.serialize('i18next', lng)
```
jamuhl: This should be not needed:

You could enable cookie on browser language detector too: https://github.com/i18next/i18next-browser-languageDetector#detector-options

`caches: ['localStorage', 'cookie'],`


add it here: https://github.com/i18next/react-i18next/blob/master/example/nextjs/i18n.js#L28

example/nextjs/server.js
```diff
@@ -23,17 +23,12 @@ const serverSideOptions = {
...
-if (enableSubpaths) {
-  serverSideOptions.detection = {
+  detection: {
     order: ['path', 'session', 'querystring', 'cookie', 'header'], // all
-    lookupPath: 'lng',
-    lookupFromPathIndex: 0,
-  };
-  serverSideOptions.whitelist = allLanguages;
-}
+    caches: ['localStorage', 'cookie']
```
jamuhl: please keep the enableSubpaths code part:

```js
const serverSideOptions = {
  fallbackLng: defaultLanguage,
  preload: allLanguages, // preload all langages
  ns: getAllNamespaces(`${localesPath}${defaultLanguage}`), // need to preload all the namespaces
  backend: {
    loadPath: path.join(__dirname, '/static/locales/{{lng}}/{{ns}}.json'),
    addPath: path.join(__dirname, '/static/locales/{{lng}}/{{ns}}.missing.json'),
  },
  detection: {
    caches: ['cookie'] // default false
 }
};

// using subpaths like /de/page2 set whitelist and detection to allow path
if (enableSubpaths) {
  serverSideOptions.detection = {
    order: ['path', 'session', 'querystring', 'cookie', 'header'], // all
    caches: ['cookie'], // default false
    lookupPath: 'lng',
    lookupFromPathIndex: 0,
  };
  serverSideOptions.whitelist = allLanguages;
}
```

jamuhl:
thank you...

And don’t forget to star this repo. Make a tweet or have a look at our https://locize.com -> there are many ways to help this project :pray:

PR #556, opened 2018-10-19
fix cookie loading

mcammaert:
Cookie loading was broken…


jamuhl:
i missed that one...but if i see right that is not needed anymore

jamuhl:
remove that as it is not needed really.

thanks for taking the time...

And don’t forget to star this repo. Make a tweet or have a look at our https://locize.com -> there are many ways to help this project :pray:

PR #560, opened 2018-10-22
Fix invalid type definitions for decorators

itka4yk:
Existed type definitions for withI18n and withNamespaces don't has state generic in return type.
This patch fixes the issue.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -43,9 +43,9 @@ export interface WithI18n extends I18nContextValues {
   i18nOptions?: ReactI18NextOptions;
 }
 
-export function withI18n(): <P extends object>(
+export function withI18n(): <P extends object, S extends object>(
   Wrapper: React.ComponentType<P & WithI18n>,
-) => React.ComponentType<P>;
+) => React.ComponentType<P, S>;
 
 export interface WithNamespaces extends WithI18n {
   tReady: boolean;
@@ -64,9 +64,9 @@ export interface WithNamespacesOptions extends ReactI18NextOptions {
 export function withNamespaces(
   namespace?: string,
   options?: WithNamespacesOptions,
-): <P extends object>(
-  component: React.ComponentType<P & WithNamespaces>,
-) => React.ComponentType<P>;
+): <P extends object, S extends object>(
+  component: React.ComponentType<P & WithNamespaces, S>,
+) => React.ComponentType<P, S>;
 
 export const translate: typeof withNamespaces;
 


index.d.ts
```diff
@@ -43,9 +43,9 @@ export interface WithI18n extends I18nContextValues {
   i18nOptions?: ReactI18NextOptions;
 }
 
-export function withI18n(): <P extends object>(
+export function withI18n(): <P extends object, S extends object>(
   Wrapper: React.ComponentType<P & WithI18n>,
-) => React.ComponentType<P>;
+) => React.ComponentType<P, S>;
```
schettino: I'm not sure if we need this, as there is no upstream changes that is going to affect an internal state type definitions.

itka4yk: Actually it will be correct definition for React components, because they can have state.

For example this is output of VSCode:

Code: 
```Typescript
import * as React from 'react';
import { withI18n, TransProps } from 'react-i18next';

interface IState {
  something: string;
}

@withI18n()
class Sample extends React.Component<any, IState> {
  render() {
    return null;
  }
}

export default Sample;
```

Error:
```
Unable to resolve signature of class decorator when called as an expression.
  Type 'ComponentClass<any, any>' is not assignable to type 'void | typeof Sample'.
    Type 'ComponentClass<any, any>' is not assignable to type 'typeof Sample'.
      Type 'Component<any, any, any>' is not assignable to type 'Sample'.
        Types of property 'state' are incompatible.
          Type 'Readonly<any>' is not assignable to type 'Readonly<IState>'.
            Property 'something' is missing in type 'Readonly<any>'.
```

schettino: Oh I see, as decorator is still an experimental feature, supporting it could bring some overhead, when we can simplify things here. ComponentType only accepts one argument by the way. I've tested your example and if you use it without the decorator, it works as expected.

index.d.ts
```diff
@@ -43,9 +43,9 @@ export interface WithI18n extends I18nContextValues {
   i18nOptions?: ReactI18NextOptions;
 }
 
-export function withI18n(): <P extends object>(
+export function withI18n(): <P extends object, S extends object>(
```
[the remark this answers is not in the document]
itka4yk: Hmm, as far as I've tested, everything is ok without state.

index.d.ts
itka4yk: Yeah, experimental, but very useful, especially in typescript. IMO correct supporting of decorators will help in future, when it becomes not-experimental.

And by the way, if ComponentType accepts only props, then how can we use ```withI18n``` with stateful components?

index.d.ts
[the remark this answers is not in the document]
schettino: works if you don't pass Props as param to the generic

index.d.ts
schettino: so yeah, the work-around-code we may need can be the pretty much the amount of code we have for the typings right now, and even so it could bring breaking changes for those who is not using decorators. As there is just a couple of hoc's here, it might be better to stick with the expected usage and as soon as the decorator feature gets stable within typescript, you can ping me here so we can start on that

index.d.ts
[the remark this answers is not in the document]
itka4yk: What do you mean?

index.d.ts
itka4yk: So maybe it's better to change types to ```any```,  as far as typings are not correct for every use cases?

Tsury:
@itka4yk I'm suffering from this issue myself. Were you able to work around it?

Issue #563, opened 2018-10-22
Module import breaks on Typescript


matteoferigo:
**react-i18next v8.1.0**
**typescript v3.0.3**

```
/node_modules/react-i18next/index.d.ts
(2,8): Module '"/node_modules/@types/i18next/index"' has no default export.
```

Index file should import `i18next` module with ```import * as i18next from 'i18next'``` syntax

matteoferigo:
`i18next` is a dependency of `react-i18next`, that's why TS tries to import `@types/i18next` types

I created a pull request to fix this problem https://github.com/i18next/react-i18next/pull/564

jamuhl:
@matteoferigo you might fill the gap: https://github.com/i18next/i18next/issues/1115

but be aware: `import * as i18next from 'i18next'` would be the wrong thing todo -> use the default export in any case -> import * won't be a valid instance being usable down the road

schettino:
@matteoferigo it's a devDependency and I could not reproduce this even with 3.0.3. Which ide are you using? It might be a good thing to check if  `@types/i18next` is installed on your project. We could clarify this on the docs, what do you think about it?

jamuhl:
@matteoferigo what i mean is

https://github.com/i18next/i18next/blob/master/src/index.js#L3

the default export is a real i18next instance

`import * as i18next from 'i18next'` gives you something very similar which works in most cases - but will kick your ass very bad when eg. using that in some i18n.js file re-exporting that and passing that around eg. in the next.js sample!!!

```js
import * as i18next from 'i18next'

i18next.init()

export default i18next;
```

now injecting that by provider or some other component taking it as prop you will get issues like `i18next.options is undefined`

As on index.js we expose just some bound functions so one can use 

`import { init } from 'i18next'` or `import { t } from 'i18next'` -> convinient but dangerous if not knowing what it is.

matteoferigo:
@jamuhl absolutely you are right. But I don't want to change the package exports, I simply want to change the type definition (index.d.ts)

I'm not totally sure, but it seems like that the `import from 'i18next'` in `index.d.ts` is referencing to `@types/i18next/index.d.ts`: imported properties are interfaces and types, and in `@types/i18next/index.d.ts` there is no `export default`, it's simply exporting the i18next namespace (and that's why tsc is yelling)

So my PR is preserving package default export, but it's simply solving the type definition mistake
Does it work for you?

jamuhl:
[...]

PR #564, opened 2018-10-22
Solved issue #563

matteoferigo:
No more troubles with Typescript 3


index.d.ts
```diff
@@ -1,13 +1,10 @@
 import * as React from 'react';
-import i18next, {
```
schettino: the only thing needed was importing `i18n` here

schettino: I'll close this as it's fixed on another PR. Thanks for pointing that out

Issue #565, opened 2018-10-23
withNamespaces definition accepts only a string as the first parameter, and not also an array of strings or function


lbevilacqua:
I updated to the last version and, beside the problem mentioned in #563, which I solved easily by dropping the @types package from my package.json, I found out that the declaration of the withNamespaces HOC is 
```
export function withNamespaces(
  namespace?: string,
  options?: WithNamespacesOptions,
): <P extends object>(
  component: React.ComponentType<P & WithNamespaces>,
) => React.ComponentType<P>;
```
instead of 
```
export function withNamespaces(
  namespace?: string | string[] | ((props: P) => string| string[]),
  options?: WithNamespacesOptions,
): <P extends object>(
  component: React.ComponentType<P & WithNamespaces>,
) => React.ComponentType<P>;
```
for the `namespace?` parameter.

Issue #562, opened 2018-10-22
HOCs need to be explicitly typed


nmenke:
Hi,

i have updated to the newest version of react-i18next today. I have noticed that the repository now contains typings. So for example the following definition:

```
export function withNamespaces(
  namespace?: string,
  options?: WithNamespacesOptions,
): <P extends object>(
  component: React.ComponentType<P & WithNamespaces>,
) => React.ComponentType<P>;
```

With this definition i need to say `withNamespaces("default")<Props>(component)`. Otherwise the compiler complains for missing Props like tReady.

```
type Omit<T, K> = Pick<T, Exclude<keyof T, K>>;
type Subtract<T, K> = Omit<T, keyof K>;

export function withNamespaces(
  namespace?: string,
  options?: WithNamespacesOptions,
): <P extends WithNamespaces>(
  component: React.ComponentType<P>,
) => React.ComponentType<Subtract<P, WithNamespaces>>;
```

This definition removes those types supplied by the HOC and you can use following statement: `withNamespaces("default")(component)` without compiler errors. It would be great if those typings would be used for every HOC.

schettino:
PR coming. @jamuhl want to patch the changes right away or give it sometime to see if something else comes up in this initial phase and then release the possible changes all at once?

jamuhl:
@schettino i added you as contributor with write access - feel free to merge in changes as needed - and ping me if there is a state we should publish again...

schettino:
@jamuhl sounds good, thanks

PR #567, opened 2018-10-23
Improve typings for HOC's and fix imports

schettino:
Closes: #565 #563 #562 #568 #571 

 - Fixed namespace type def and for when a extractor function is provided
 - Removed unneeded default imports
 - Improved types for hoc's


jvgreenaway:
These changes look good to me

schettino:
@jamuhl whenever it's possible, we can release this patch

jamuhl:
published in react-i18next@8.1.2

thanks a lot for the work done here

Vrajmas:
Thanks, best regards

Issue #574, opened 2018-10-24
nextjs example page3 serving wrong language during ssr


Enalmada:
I am trying to get started with nextjs and react-i18next and hitting a problem that has me completely blocked.  I would like to pass "t" down from my _app.js to pages without wrapping each individual page in "export default withI18next()".    

essentially in _app.js
```
<NamespacesConsumer {...pageProps} ns="common" i18n={(pageProps && pageProps.i18n) || i18n} wait={process.browser} >
          {t => ( <Component t={t} {...pageProps} /> )}
</NamespacesConsumer>
```

Then pages are pure components:
```
export default const Page2 = ({t}) => (
        <h1>{t("welcomePage2")}</h1>
);
```

To reproduce:
Start react-i18next/example/nextjs
switch language to de
navigate to page 3: http://localhost:3000/de/page3
do a browser refresh of the page and note content is in en first and flickers to de
view source - the h1 from _app.js wrapped with "NamespacesConsumer" is in en not de

This is essentially my bug...the `<h1>{t('common:integrates_react-i18next')}</h1>` in _app.js seems correctly wrapped in NamespacesConsumer and should be rendering proper locale regardless of what pages under are doing.

Could you please help fix the example so that the page3 example renders the _app.js content in the correct language while keeping it a pure component?

jamuhl:
did you move all translations to common.json?

as you only load that in the _app.js

`<NamespacesConsumer {...pageProps} ns="common" `

jamuhl:
not even sure this is the way it works i guess the flow with app.js -> i guess that gets wrapped magically were in reality the page is the outer component (as pageProps come from the page)...

but honestly i'm far from an expert in next.js - so most parts are magic to me

Enalmada:
@jamuhl I really appreciate you taking the time to help me out despite your limited experience with it.    In my project I am ok with everything being in common btw as long as ssr works in pure page components.  Fortunately this issue is very easy to reproduce...fire up the nextjs example, click button to change to de, click page3 link, do a browser refresh on page and you will see the flicker and page source being in english.   I feel like putting the `export default withI18next(['common'])(MyApp)` in _app should work and fix this...I assume there is some magic that withI18next is doing to the pages that NamespacesConsumer isn't...otherwise the t("") call from _app.js would look fine on page3 refresh.  Perhaps the circular dependency cause is worth looking into?

jamuhl:
[...]

PR #580, opened 2018-10-26
nextjs example page3 ssr rendering fix

Enalmada:
This fixes the nextjs example page 3 which currently only shows default language during ssr.
Fixes https://github.com/i18next/react-i18next/issues/574

### DIFF
--- a/example/nextjs/pages/_app.js
+++ b/example/nextjs/pages/_app.js
@@ -31,8 +31,22 @@ if (enableSubpaths) {
 // Warning: Did not expect server HTML to contain a <h1> in <div>.
 // not sure - did neither find something wrong - nor seems the warning to make sense
 export default class MyApp extends App {
+
+  static async getInitialProps({Component, router, ctx}) {
+    let pageProps = {};
+
+    if (Component.getInitialProps) {
+      pageProps = await Component.getInitialProps(ctx);
+    }
+
+    // initialI18nStore and initialLanguage need to be passed into NamespacesConsumer
+    // for any page that is not wrapped with withI18next to enable ssr localized rendering
+    let i18nProps = i18n.getInitialProps(ctx.req, "common");
+    return {pageProps, i18nProps};
+  }
+
   render() {
-    const { Component, pageProps } = this.props;
+    const { Component, pageProps, i18nProps } = this.props;
 
     return (
       <Container>
@@ -41,6 +55,8 @@ export default class MyApp extends App {
           ns="common"
           i18n={(pageProps && pageProps.i18n) || i18n}
           wait={process.browser}
+          initialI18nStore={i18nProps.initialI18nStore}
+          initialLanguage={i18nProps.initialLanguage}
         >
           {t => (
             <React.Fragment>


jamuhl:
how does this fix Page3 -> Page3 is a page testing all works without having that page extended...like having an untranslated page (for what reason ever)

Enalmada:
If you run the example, you will see that is not quite true.  _app.js functions as a site wide decorator and has a `<NamespacesConsumer>...{t('common:integrates_react-i18next')}...</NamespacesConsumer>` at the top making it show up on all pages.  Because page 3 has no HOC `withI18next` on it, the  `t('common:integrates_react-i18next')` fails to render in ssr.  This fixes that.

jamuhl:
ok agree...but now having initialStore on both the _app.js and pages introduces a lot of problems...believe me

so better just delete page 3 for the sample

Enalmada:
I believe you...no one knows this stuff like you!  Having the page 3 example made all the difference for me as an end user as I don't want to use withI18Next HOC (the full story is that I am using reason-react and HOC in reason is [technically possible](https://medium.com/@_gabrielrubens/binding-a-react-hoc-to-reasonml-children-as-function-d2688e4edaae) but not something I have not figured out yet).  Perhaps there is a compromise to be made...perhaps we pull the change language button out of _app.js and put it on page 1 and page 2...page 3 can be the example for using NamespacesConsumer without withI18next?

jamuhl:
Page3 is really just plain test - so best is removing it.

The problem i see with this here...we add more complexity - so sure it semi solves the issue for you when removing all hocs - but breaks the sample for others having hocs...

so adding reason-react into the mix just adds to the complexity...so my suggestion would be make a sample of it for others...make a blog post and share your knowhow using reason-react, react-i18next with next.js so this knowledge can be found by others. That would be 🔥

Enalmada:
Page 3 isn't just a plan test...it is an example of choosing to use Component and passing t down to children...a very real thing many users will feel is best to do.  It is common to be able to choose between using HOC or Component....you even recommend passing t down to children in several tickets.  The moment I tried this on my project I got broken ssr rendering.  I couldn't disagree more with proposing to delete pages of the example just to hide and avoid fixing the bug that ssr rendering doesn't work unless you use the HOC.

I have done my best to fix the currently broken example with this pull request.  If you are unhappy with the proposal, then on behalf of the community of people that want to use a pure component method with ssr I ask you to consider which one of these options is easiest:
- fix circular dependency error when you add withI18Next to _app.js
- fix NamespacesConsumer so it doesn't require initialI18nStore/initialLanguage to work with ssr

jamuhl:
Disagree partially...the current sample shows using hocs -> so it should be consistent for that...you suggest doing a sample without hocs - which is another option of solving this.

i advocate not using hocs on inner elements to often -> best using it on larger container components and not on each link, li whatever custom tiny element in a eg. GUI component lib.

So i say both solutions are right - but are exclude each other? agree?

jamuhl:
1) fix circular dependency error when you add withI18Next to _app.js - not aware of that but seems to be the problem of how _app.js works
2) fix NamespacesConsumer so it doesn't require initialI18nStore/initialLanguage to work with ssr the point is ONE and just one element can set initialI18nStore and initialLanguage either the _app.js or each page for itself -> as not all project use _app.js the sample goes for page

jamuhl:
@isaachinman no pressure - but would be happy about a 3rd opinion on this

Enalmada:
I see in issue history you have solved other circular dependencies...any chance this one might be similar enough it might be a quick fix based on your previous experience?  If you can't easily fix it then we can create separate ticket because based on your input, it sounds like it wouldn't be a good idea to do withI18Next("common") in _app.js and then in each individual page do withI18Next("page") as an override.   

> ONE and just one element can set initialI18nStore and initialLanguage either the _app.js or each page for itself

That makes sense. Based on your feedback in #574, NamespacesConsumer without HOC requires that you pass in initialI18nStore and initialLanguage to work with ssr.  The current example is broken because it doesn't do that and I have done my best to fix.  Is there some way we could modify the example to demonstrate to new users that just want to use NamespacesConsumer how it needs to be to work with ssr?

Enalmada:
page 3 isn't confusing, it exposes a real issue in react-i18next.  The server only ssr renders default language from i18n elements in _app.js.  This PR is implementing the workaround jamuhl gave me in #574 in my attempt to help users avoid the same pain I experienced.  If a user for some reason has i18n elements in their _app.js, which is a very real possibility, then this PR is the only known way of avoiding  putting withI18Next on every page of their site, including pages that don't otherwise need it.  I hit the problem trying to get i18n working on my own site and only discovered page 3 had the problem after the fact when trying to copy it to make the smallest repeatable test case for you knowledgeable experts.  

other potential solutions if this PR is really unacceptable:
- remove i18n element from _app.js 
- make withI18Next work on _app.js in a way that it isn't bad if users also have withI18Next() in child pages.  jamuhl seemed to indicate with strong language that it is better to pass t down than have withI18Next() on children elements.  Do you two disagree or do I misunderstand?  Sorry I am new to this.
- delete page 3 and document somewhere that all new pages must have withI18Next() or stuff in _app.js will break
- put withI18Next on page 3 with notes that withI18Next must be on every page if you have i18n in _app.js.
- improve NamespacesConsumer so that you don't need to pass language and store for SSR to work.  What is the point of NamespacesConsumer if you need withI18Next() on every page?  I am sorry if that is a silly question.  It seems like its whole point is to be an alternative to withI18Next and it is broken in ssr without passing the parameters this PR is attempting to educate users that want to do that how to do it.

jamuhl:
@isaachinman Regardless of how/where you use the HOC, we do have an issue of tree traversal, and currently do not discover nested namespace dependencies.

=> that was solved with this change -> https://github.com/i18next/react-i18next/blob/master/example/nextjs/lib/withI18next.js#L20

and https://github.com/i18next/react-i18next/blob/master/example/nextjs/lib/withI18next.js#L35

It will report namespaces consumed deeper in the tree (using withNamespaces, NamespacesConsumer on nested components)

jamuhl:
@Enalmada no _app.js renders in same language as all other t calls on pages.

the case we discussed is you liked to not use hocs but pass the `t` function from _app.js to your pages -> therefor not having a withI18next HOC handling initialState you need to add that to the _app.js
-> this PR mixes it

again Page3 as is in this sample neither gets `t` from props nor by hoc -> so its just a plain page not using i18n at all (beside the one title coming from shared layout in _app.js)

jamuhl:
@Enalmada now i see your problem page3 does not translate the title in _app.js (didn't get that from reading the issue -- sorry for that -- wasn't clear enough to me 🤕)

- remove i18n element from _app.js

=> works at least on my machine https://github.com/i18next/react-i18next/blob/master/example/nextjs/pages/_app.js#L47 is in the same language as everything else

- make withI18Next work on _app.js in a way that it isn't bad if users also have withI18Next() in child pages. jamuhl seemed to indicate with strong language that it is better to pass t down than have withI18Next() on children elements. Do you two disagree or do I misunderstand? Sorry I am new to this.

=> had issues with that - so i gone for the NamespacesConsumer usage (might be some internal of next.js how it includes _app.js into the render tree)

- delete page 3 and document somewhere that all new pages must have withI18Next() or stuff in _app.js will break

=> seems the solution to me - at least for this sample (until we won't change it from using hocs to do everything in _app.js -> which is no real option neither as not everyone uses _app.js)

- put withI18Next on page 3 with notes that withI18Next must be on every page if you have i18n in _app.js.

=> same as above Page3 has no real added value so removing it is better

- improve NamespacesConsumer so that you don't need to pass language and store for SSR to work. What is the point of NamespacesConsumer if you need withI18Next() on every page? I am sorry if that is a silly question. It seems like its whole point is to be an alternative to withI18Next and it is broken in ssr without passing the parameters this PR is attempting to educate users that want to do that how to do it.

=> withI18next is no official react-i18next component it's a convenient addon to solve this...yes we could make the same as for the withNamespaces HOC for the NamespacesConsumer render prop -> but will it add value? i guess no -> will just add to complexity

jamuhl:
@Enalmada @isaachinman i removed Page3 for now.

jamuhl:
@Enalmada sorry for the confusion caused by me...really didn't payed to much attention to page3 -> removed that now as it seems to cause more problems.

adrai:
sorry was not my intension

Enalmada:
There was nothing wrong or confusing with the example code page 3.  It is completely valid to have i18n elements in _app.js and pages without withI18next on them.  Page 3 just exposes the bug that happens when users do this.  You are choosing to hide the issue that that NamespacesConsumer, if used alone in _app.js and t passed down, requires store and language parameters passed in for ssr to work in nextjs rather than address it.  I get that it is your opinion that withI18next() is better, but withI18next() doesn't work in _app.js due to another bug.   NamespacesConsumer is the only way to get i18n working in _app.js.  Users are going to try and use NamespacesConsumer to pass down t just like the example in _app.js implies they can do it.  They will do this and not realize it is always broken in ssr because the visible screen is in the correct language.  They are going to have poor seo because the pages will always be indexed in the default language.  The workaround to enable SSR with NamespacesConsumer is quite easy, just pass in store and language as this PR shows how to do...but users will not know how to do that if it is not in the example. 

How do you guys feel about at least a comment in _app.js above NamespacesConsumer noting that if you choose to use NamespacesConsumer to power i18n in _app.js without withI18Next on a page, you must pass in store and language parameters for SSR to work?

jamuhl:
Not page3 was confusing -> i was confused with the issue - as i wasn't aware with the problem (as _app.js was introduced later than page3 ---> so i never loaded page3 as the first page load).

There is no right or wrong. This sample prefers using withI18next on each page. It's a design decision for this sample. So the sample goes for the tradeoff having each page wrapped in a withI18next.

Just to make clear...the sample without Page 3 works like needed...content in _app.js gets translated on server (turn js off).

Doing what is in the PR will break the initialStore setting - point if you set it like in the PR it will only pass down common - but not namespaces requested deeper down the tree. We can delete this sample and deleget setting initialStore and initialLanguage in _app.js -> than devs can complain they don't use _app.js and the sample does not work if removing it.

Like said before i have nothing against a comment like you describe...even a link to a medium post or sample removing withI18next and doing all initial loading in _app.js - removing the benefit of splitting translations into multiple files to avoid loading all the translations on inital page...-> no translation splitting no benefit.

Enalmada:
Sure, the sample works now because the bug users are likely to hit eventually has been deleted from example rather than addressed.   I have no problems with the design decision.  I have problems with the following because it happened to me personally:
- The moment a user with i18n content in _app.js and i18next() on every page following the design decision happens to add a page to their project without withI18next() on it for some reason, their ssr will break on that page, likely without them ever realizing it. 
- A subset of users (like me) will that see NamespacesConsumer in the current example _app.js is powered by t and may attempt to pass down that t to child components and it will appear to work, but ssr will be broken for them and they may never realize that because the consumer renders correctly.  

I was hoping this PR would reduce the chances of other users experiencing broken ssr for any reason and in situations where it is almost impossible to tell it is happening unless you are looking at page or have js turned off.

I believe it is totally ridiculous to resort to a medium post about how NamespacesConsumer is broken in ssr right now.

jamuhl:
Like said...what you like todo is fixing a potential problem that is introduced by leaving what the sample shows with another issue. Page3 will work but all pages with withI18next and withNamespaces deeper down the tree will fail. (Try to understand how initial state is handled in react-i18next -> you can't set that twice -> first wins)

At this point i really give up to explain why your PR which should be a fix breaks the sample more. It's an opinionated sample....not a it will solve every devs code independent how hard he changes the sample.

Enalmada:
I think you might be getting too hung up on my original proposal...I understand how it is bad.  Please forget about original proposal and just focus on how both issues I pointed out above above are valid issues users could hit.  I only want to reduce the chances of either of them happening to real users however you think best.  I am open to anything, even just some notes in the example warning users with all the valuable info you have provided in this thread.

isaachinman:
@jamuhl Nice to hear about the tree traversal with `reportNS`. If that is the case, in my humble opinion, all components that contain translated content (be it pages themselves or smaller components) should be decorated with the HOC and specifically declare their individual dependencies.

At certain points, it is going to be impossible to build a practical example without taking an opinionated approach. In my experience, this decorator/HOC pattern most accurately reflects enterprise-level best practices.

@Enalmada Thanks also for your concern and efforts to help here. It's nice to see the community is actually using the example, and we certainly want it to be as helpful/clear as possible.

Also in my humble opinion, it's most likely time to close/lock this though, as it's not a very clear discussion and has gotten off topic.

Enalmada:
@isaachinman There are a few cases I have identified that users can easily cause ssr to break.  Ways users are likely not to notice because client side looks right....which is probably the worst kind of problem possible.  Ways I feel are legitimate uses of the code.  If there are ways users can easily get broken ssr, they should either be documented or fixed.   

Perhaps we are just having a language issue and what you are really trying to say is that you just don't feel it is worth documenting or fixing the issues I have identified.

jamuhl:
@isaachinman yes...i try most time to base the samples on my experience doing large-scale applications means adding spitting translations into multiple files -> could go to make it the simplest to start to get more retweets, demos done as those are dead simple -> just going for the real thing it helps no one as it's to limited and simplifying the real problems.

@Enalmada it's a sample - when devs break it and don't proper test SSR i feel sorry but at some point i need to have a life -> see things like this: https://github.com/i18next/react-i18next/issues/582 i get such stuff almost daily here or on other repos or to my email. I maintain i18next - i can't teach everyone doing everything the right way (i'm far from a next.js, gatsby, react-native pro - but doing samples you need understand both those frameworks coming with their up and downs and react-i18next ---> that's what makes it harder than it seems.)

I understand your worries...and fully agree on getting it documented. there is https://react.i18next.com feel free to start adding a next.js guide there. I personally burned out for this week -> having done omi-i18n, react-i18next with hooks and closed endless issues. It's not that i don't have a plan for add some guidance for next usage...but my personal backlog currently has other priorities like doing a sample for gatsby to help those devs too. Then i like to reach out to all the different framework maintainers where multiple implementations exist (angular, vue) to get a consolidated solution - no more choosing from n possible i18next candidates for that framework.

Guess we all agree - we have the same in mind and good intentions. Let's close this for now.

Would be happy to see help with documentation on a next.js guide...

isaachinman:
@jamuhl Thank you for all the time you've put into this thread. I was quite busy in the last week and distracted, so my responses weren't super helpful. I had an hour today to actually look at this.

In general, communication on this issue wasn't as good as it could have been. It seems like the problem finally was clarified to: **if you render a React tree that does not contain the `withI18next` HOC, the default language will not be correctly selected during SSR.** It turns out this is a trivial issue. The problem in this issue was communication.

So, to close this issue once and for all:

>  It is completely valid to have i18n elements in _app.js and pages without withI18next on them.

I think this is where the confusion began. In the context of this example, it _is not_ valid to have translated components which are not wrapped by the `withI18next` helper. In my opinion (_opinion_ being the key word), it's not appropriate to have any translated content in `_app.js` at all. If you want some global UI, make that into a `<Header />` component, and wrap it with the `withI18next` decorator so that each individual component's namespace deps are clear.

> users can easily cause ssr to break. Ways users are likely not to notice because client side looks right

In general, SSR is an advanced topic and the working assumption is that all developers attempting SSR will understand the difficulties involved and be able to debug. SSR becomes even more complicated in the context of localisation. For the last time, we are not providing a perfect solution. We are providing an example which you can use as inspiration in your own professional work.

Regardless of all of that, I do think @Enalmada's points are valid, and perhaps we shouldn't rely on the HOC to detect initial lang. I've created #592 to fix this. @jamuhl Please note that I added `page3` back in to demonstrate the fix working, but that commit can easily be removed if you'd rather keep it out. We can also most likely simplify the HOC and/or other parts of the codebase now, too.

If it is helpful to add something to example's README, just let me know - I would be happy to improve documentation in whatever way possible.

@Enalmada If for some reason, one of your questions remains unanswered or unsolved, please open a new issue with a clear and concise description, and we'll fix it.

Issue #590, opened 2018-10-28
FEEDBACK WANTED FOR NEXT MAJOR VERSION: using HOOKS and SUSPENSE


jamuhl:
We are rather amazed at how much smaller and easier react-i18next feels using hooks and suspense.

=> Early Feedback would be very welcome 🙏

A little introduction: https://react.i18next.com/experimental/using-with-hooks

sample code: https://github.com/i18next/react-i18next/tree/master/example/react-hooks

new source code: https://github.com/i18next/react-i18next/tree/master/src/hooks

schettino:
at first glance, it seems very straightforward. I'd say, however, to use `useTranslation` instead of `useT`. Different from the `t`, which is enough to express what we want, `useT` might lack a little of that expressiveness, what do you think?

jamuhl:
@schettino totally agree will update that on the next release to `useTranslation` and `withTranslation` - no other hook use cryptic short namings - so we shouldn't do either

jamuhl:
Yes, not being able to pass a function, dan already mentioned that this is a problem on twitter - i guess that will be sorted out at some point (for now it's an "acceptable" workaround).

Yes, there is not much functionality removed (beside not using context api, removing the render prop - which could easily be added in userland if needed) - just need to figure out the SSR part again. Not yet sure if by adding an explicit hook, Component or by just passing a second argument to the `useT` hook containing the props (initialStore, initialLanguage). `reportNS` also is not in right now.

jamuhl:
@isaachinman throwing the Promise -> that is how Suspense works under the hood...it catches and if it's a promise -> showing the fallback. When all promises resolve it goes back and renders the actual content...a little weird but seems to work 🙄

isaachinman:
Ah, I didn't see that tweet. As long as they're aware of it, and are treating it like a bug, no problem. Also strange to hear that about suspense. I watched Dan's talk awhile back but haven't gotten hands-on with it yet. Throwing a promise seems like a really weird API, but I guess good to know!

Maybe you can explain a bit more about what we need to get SSR working again? In my opinion a minor-bump-refactor is not worth it if it means dropping crucial functionality like SSR. Is it your plan to get SSR working before merging this? It sounds like you have a few possible approaches in mind. We can discuss if it is helpful to you.

jamuhl:
@isaachinman not yet sure if there also will be some overhaul on next.js itself...but for sure SSR will need to work for v10.

[...]

PR #592, opened 2018-10-28
Fix default lang bug when React tree doesn't contain withI18next

isaachinman:


jamuhl:
thank you for looking deeper into it...really appreciate your help.

personal i'm looking into the 16.7 release of react -> guess will bring some changes to next.js too

isaachinman:
Yup, it probably does. I haven't looked at 16.7 yet, but I might get a chance to this week. Just let me know if you need anything!

jamuhl:
prepared a react-i18next version based on hooks and suspense (no support for SSR yet) -> if you ever into it you might have a look: https://github.com/i18next/react-i18next/issues/590

isaachinman:
Nice! I will take a look sometime this week.

Enalmada:
@isaachinman When I run the latest nextjs example from master, clicking any link appears to do a full page change and flicker.

Issue #594, opened 2018-10-30
Comply with react strict mode


JimmyLaurent:
React strict mode complains about the use of "componentWillReceiveProps" in "I18nextProvider".

Can we can replace it with "getDerivedStateFromProps" as suggested by the warnings ?

Thanks

jamuhl:
Sure...are you open for a PR? Else i will add the change in the next version. Guess not to urgent to get patched.

JimmyLaurent:
Done.

I had to mirror the i18n prop in the state, though.

And you're right, it's not urgent.

jamuhl:
Thank you...will take a look later today

jamuhl:
should be solved with 022b9aa

published on npm in react-i18next@8.3.6

If you like this module don’t forget to star this repo. Make a tweet, share the word or have a look at our https://locize.com to support the devs of this project -> there are many ways to help this project :pray:

PR #596, opened 2018-10-30
Comply with react strict mode

JimmyLaurent:
Replace "componentWillReceiveProps" with "getDerivedStateFromProps" in "I18nextProvider".

Fixes: #594

### DIFF
--- a/src/I18nextProvider.js
+++ b/src/I18nextProvider.js
@@ -3,6 +3,12 @@ import { I18nContext } from './context';
 import { initSSR } from './utils';
 
 export class I18nextProvider extends Component {
+  static getDerivedStateFromProps(nextProps, prevState) {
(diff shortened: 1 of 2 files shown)


jamuhl:
As it's only to show a warning if someone actively replaces the i18n instance - what do you think of dropping that warning? makes it simpler - and that warning i don't know how many people run into this honestly?

JimmyLaurent:
I totally agree, it feels like an edge case.

Do you want a new PR ?

jamuhl:
If you like and not takes to much time...else i will just remove it next time i got it open.

jamuhl:
should be solved with https://github.com/i18next/react-i18next/commit/022b9aa5df630cb413c3e7746ecbf7569e638a9f

published on npm in react-i18next@8.3.6

PR #598, opened 2018-10-31
Plugs i18next-react-native-async-storage in the React native example

0xClpz:
Hi! 

[I wrote a plugin](https://github.com/0xClpz/i18next-react-native-async-storage) for your library that saves the user's language in React native's async storage and I thought it'd be nice to use it in the example since this is something you (probably) want to do in any of your react native apps!

### DIFF
--- a/example/reactnative-expo/js/i18n.js
+++ b/example/reactnative-expo/js/i18n.js
@@ -1,19 +1,21 @@
 import i18n from 'i18next';
 import { reactI18nextModule } from 'react-i18next';
 import Expo from 'expo';
+import AsyncStoragePlugin from 'i18next-react-native-async-storage'
 
-// creating a language detection plugin using expo
-// http://i18next.com/docs/ownplugin/#languagedetector
-const languageDetector = {
-  type: 'languageDetector',
-  async: true, // flags below detection to be async
-  detect: (callback) => { return /*'en'; */ Expo.DangerZone.Localization.getCurrentLocaleAsync().then(lng => { callback(lng.replace('_', '-')); }) },
-  init: () => {},
-  cacheUserLanguage: () => {}
+// creating a language detection fallback using expo
+const detectUserLanguage = (callback) => {
+  return Expo
+    .DangerZone
+    .Localization
+    .getCurrentLocaleAsync()
+    .then(lng => {
+      callback(lng.replace('_', '-'));
+    })
 }
 
 i18n
-  .use(languageDetector)
+  .use(AsyncStoragePlugin(detectUserLanguage))
   .use(reactI18nextModule)
   .init({
     fallbackLng: 'en',
--- a/example/reactnative-expo/package.json
+++ b/example/reactnative-expo/package.json
@@ -10,10 +10,11 @@
   },
   "dependencies": {
     "expo": "^30.0.1",
-    "react": "16.3.1",
-    "react-native": "https://github.com/expo/react-native/archive/sdk-30.0.0.tar.gz",
     "i18next": "11.9.0",
+    "i18next-react-native-async-storage": "1.0.2",
+    "react": "16.3.1",
     "react-i18next": "8.0.6",
+    "react-native": "https://github.com/expo/react-native/archive/sdk-30.0.0.tar.gz",
     "react-navigation": "2.17.0"
   }
 }


0xClpz:
Right it’d make sense to store that in your external data store then persist your data store, I just built this for one specific project but I guess in most cases you wouldn’t do it this way 👍 anyway I’d be happy to hear your suggestions for the plugin even if it doesn’t make it into react i18next ! 

Thanks for the quick reply !

isaachinman:
Some very quick thoughts:

1. This absolutely, 100%, should be independent of Expo.
2. Would be nice to see optional use of `NativeModules.I18nManager` for default locale detection.
3. Definitely need to be able to manually specify the `AsyncStorage` key.

I didn't see your use of `Expo` until a second look through just now. That definitely disqualifies the plugin from use in any example, at least until you remove Expo as a dep.

0xClpz:
Mmh I might be wrong but it doesn’t depend on expo ? The fallback mechanism in the ready is an example of what you could do if nothing is in the Async storage but the actual lib never makes uses of expo

Mhm yeah I guess it could be nice to have default language detection built in along with the storage 

I’ll add a way to pass a custom key for the async storage !

Thank you very much for the code review !

isaachinman:
Apologies, it is just in your example I guess. The point is that you can get default locale from RN core directly, so there's no need for that, and you might as well build it into your plugin directly.

If you want any help/review of your project in the future, you can request a review from me on a PR, or just write @isaachinman. Good luck!

jamuhl:
@isaachinman @0xClpz hm using Expo for detection was in the sample for long time -> the idea most starting with reactnative will start with Expo (create reactnative app). So this plugin - i honestly not yet had a look at the code - would persist the language too if the UI would allow calls to i18n.changeLanguage? right? If so, i guess we could add this with some hints on persisting that using mobx, redux is another option (if not needed to hook in redux i think that is a benefit for the future)

jamuhl:
@isaachinman @0xClpz might be less about this PR but in general at some point i need to update the reactnative sample i guess...like @isaachinman mention we directly can pull the device language without the use of Expo.

0xClpz:
Yeah I based the detection code in my example on the example found in this repo. There's indeed no need to hook into redux to persist the language in my example which makes it a lot simpler to persist the data from the lib I guess ?

isaachinman:
@jamuhl I haven't looked through the RN example yet, but if you want to open an issue to change that Expo dep to RN core, I would be happy to take care of it. In general I would be very wary of including Expo in any examples, or encouraging new devs to go down that route.

@0xClpz Sorry for the confusion, I didn't realise you took the example from `react-i18next`.

jamuhl:
@isaachinman agree. The question still is - if we like to add that languageChange "persist" module provided by @0xClpz - it was added to language detectors section in i18next docs already.

As we do user triggered language change in the sample: https://github.com/i18next/react-i18next/blob/master/example/reactnative-expo/js/pages/Home.js#L21 i would say it makes sense to add this.

In the next step (after merging this) we could replace the detection to the rn way here: https://github.com/i18next/react-i18next/pull/598/files#diff-b1a8277cabe82ae35e0471d497990060R7

isaachinman:
My initial impression is that persistence to `AsyncStorage` belongs squarely in user land, as nearly every application will take a different approach to that.

If you _do_ want to include persistence in the example, I would then say that it should be the simplest key/value implementation possible, with comments explaining to users what is being persisted, and how they might extend the functionality themselves.

I guess my core point is that handling persistence is not only something that can be handled by each user for their own use case, it _should_ be handled individually, and not abstracted away.

Up to you @jamuhl - either way I'm happy to work on React Native PRs as necessary.

jamuhl:
@0xClpz hm...just thought about it...with more or less every environment we got a language detector plugin (eg. browser detector comes with tons of detection options and cache to cookie or localStorage).

Why not extending your detector to include the detection via RN -> so it can do the same as in web -> take from cache or if not provided take from device -> which would make your module a complete solution where the dev does not need to pass in the detection from device. 

So we could remove the custom detection code and just like for web use your detector?

0xClpz:
@jamuhl sounds good ! I’ll update my plugin today !

0xClpz:
@jamuhl @isaachinman Hey guys!
Sorry for the delay!
This is actually trickier than I thought.

[The method using NativeModules](https://stackoverflow.com/questions/33468746/whats-the-best-way-to-get-device-locale-in-react-native-ios) only work outside of Expo so I guess I need to detect if the user is running inside of expo and detect the language accordingly, I'll tackle it as soon as I find some time for it.

PR #613, opened 2018-11-08
Full rewrite of NextJs example

isaachinman:


jamuhl:
great job - any idea if we should move this to next.js examples or update there to only have a link pointing here?

isaachinman:
Will try to get ahold of @timneutkens and see what is possible.

Either way - let's leave it for a week or two and see if it's stable enough.

jamuhl:
@isaachinman completely agree to give the new sample some weeks to get feedback - before going either way.

timneutkens:
I'd be happy to have it link to the example here, which would make it easier for you to keep it up to date

example/nextjs/lib/withNamespaces.js
```diff
@@ -0,0 +1,25 @@
...
+
+*/
+
+import { withNamespaces } from 'react-i18next';
+
+import i18n from '../i18n';
+
+export default (namespaces = []) => {
+  i18n.nsFromReactTree = [...new Set(i18n.nsFromReactTree.concat(namespaces))];
+  return withNamespaces(namespaces);
```
timneutkens: Does this HOC copy the `getInitialProps` method when used on top level pages? Otherwise getInitialProps won't work on pages wrapped in withNamespaces

isaachinman: @timneutkens That's a great point.

The `withNamespaces` HOC from `react-i18next` is already using `hoist-non-react-statics`. I just tested this by adding a `getInitialProps` static to the `page-with-hoc` in the example, and can successfully pass props down.

@jamuhl and I were discussing, and we'd like to merge this functionality into the `react-i18next` core regardless.

isaachinman:
@timneutkens That's great to hear. Is there some way of creating a symlink for GitHub repos? Or, what's the best way to achieve that? I think it _is_ possible to nest repos, but I'm not sure if you can do that cross-organisation.

example/nextjs/lib/withNamespaces.js
jamuhl: @isaachinman yes...absolutely...if you like to provide a PR to merge it in

jamuhl: @isaachinman agree

timneutkens:
The best way to do this is just clear out the example of the next.js repo side and then link to here in the readme.

isaachinman:
@timneutkens Sure, that's fine. [This](https://blog.github.com/2016-02-01-working-with-submodules/) is what I was referring to.

timneutkens:
Yeah that'd be a mess to maintain 😄

isaachinman:
Fair enough. Do you want a PR or will you just commit to master directly?

PR #620, opened 2018-11-16
Better TypeScript support

gregbacchus:
Enables full strong typing with TypeScript

e.g.

```typescript
interface IMyRouteProp {
  id: string;
};
const Topic = withNamespaces<WithNamespaces & RouteComponentProps<IMyRouteProp>>()(({ match, t }) =>
  <h3>Foo: {match.params.id}</h3>
);
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -43,7 +43,7 @@ export interface WithI18n extends I18nContextValues {
   i18nOptions?: ReactI18NextOptions;
 }
 
-export function withI18n(): <P extends object>(
+export function withI18n<P extends object>(): (
   Wrapper: React.ComponentType<P>,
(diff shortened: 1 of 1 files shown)


schettino:
I'm not following,

using your example I can do this

```
type Props = WithNamespaces & RouteComponentProps<IMyRouteProp>;
const Topic = withNamespaces()<Props>(({ match, t }) => (
  <h3>Foo: {match.params.id}</h3>
));
```

Everything gets typed properly. Please, help me on understanding better what you're facing. I'd suggest you to open an issue describing this, what do you think?

gregbacchus:
Oh, true. Sorry, didn't get that. Best left as-is.

PR #623, opened 2018-11-20
Simplify NextJS example

revskill10:
Now we have only two files: one i18n.js for server and one i18n.js for client.


isaachinman:
@revskill10 Can you please explain your approach/reasoning?

I don't think removing a top-level `config.js` file and simply shifting the contents of several helper files into the `i18n` file solves any problem, or removes any complexity.

Helper functions are separated into distinct files within a `lib` or `utils` directory to demonstrate a separation of logic, and make our application more maintainable. It is the case that `lngPathCorrection` (for example) will be reused in countless places, including imperative NextJs routing (using the `as` prop). These functions have nothing to do with our i18n instance.

If you prefer to put everything inside the `i18n` file, that might be something for your own projects, but I'm not convinced it adds any value, and probably makes it _harder_ to understand what is going on.

Also, something like `share-config` (a modifiable configuration file) definitely shouldn't be in a `lib` dir.

The `makeConfig` function should use destructuring and should not have opinionated fallbacks like `'en'`.

Just my thoughts. Happy to hear a further explanation.

revskill10:
@isaachinman Thanks for great feedback.
The tradeoff i made is that, to limit the user api in order to make i18next work out of the box. All the user have to do is just modifying the `share-config.js`, which is just a JS object.

For your point on reusing, in this case, make a copy of that `lngPathCorrection` is not a bad idea. The benefit is that, `i18n.js` is alone standing, easier to reuse between projects (less moving things).

THere're two improvements that i could make:
- Expose only `registerI18n` for each server and client.
- Expose only `registerI18n` for both server and client.

The point i made here, is that, user only needs to edit the `share-config.js` to make things work out of the box between many projects.

revskill10:
@isaachinman Great to know. I just try out the way and the simplest API possible in order to reuse. I think this is only first step. This needs more investigation of course. So i think i'll close this PR here, and anyone could use this as a reference.

Surely i will contribute to `next-i18next` when i test it. It'll be a full-featured package !

isaachinman:
Great, I look forward to your help!

PR #624, opened 2018-11-23
Set ns as optional in Typescript definition

francescopersico:


francescopersico:
Please merge the PR and create a new release with the fix.

jamuhl:
@francescopersico do you need this urgent as a hotfix or can we release this with the next release?

PR #631, opened 2018-11-30
match the typescript definition with that of the latest version of i18next

emilgpa:


rosskevin:
awesome thanks @emilgpa!  @jamuhl any reason to delay this in a release?  I just bumped into this mismatch when updating all dependencies (we do about every three weeks).

jamuhl:
published in react-i18next@8.4.0

PR #557, opened 2018-10-20
Added Type Definitions

schettino:
_For future releases, feel free to ping me to get the typings updated_

Resolves: #550 

This PR adds the typescript type definitions.

hoc usage
```javascript
interface Props extends WithNamespaces {}

export default withNamespaces()<Props>(AddressField);
```

PR #634, opened 2018-12-04
Restore tOptions typing needed for the ICU plugin

kachkaev:
This PR restores `tOptions` prop, which was lost while migrating from DefinitelyTypes typings. See https://github.com/i18next/react-i18next/pull/557 and https://github.com/DefinitelyTyped/DefinitelyTyped/pull/29945 (file ` types/react-i18next/src/trans.d.ts`).

cc @schettino


jamuhl:
closing like you mention using values is nicer

PR #641, opened 2018-12-12
Update README.md

gavinbarron:
It's not currently obvious as to how to use the `@withNamespaces` decorator in Typescript unless you manage to find the PR that introduces that support. Putting it into the readme will save others a deal of frustration


jamuhl:
Would prefer having that in the documentation rather than in the readme: https://react.i18next.com/components/withnamespaces

jamuhl:
closing in favor of: https://react.i18next.com/components/withnamespaces#using-with-typescript

Issue #591, opened 2018-10-28
v10: hooks and suspense


jamuhl:
The plan is to provide both versions in the future:

- **v10** will be the version having the new **hooks and suspense** version only
- **v9** will be the last major version having the **current** react-i18next available (will go into maintenance mode -> only bug fixes - no new features)

We for sure will have to maintain v9 for a while.

TODO:

- [x] rename `useT` -> `useTranslation`, `withT` -> `withTranslations`
- [x] report namespaces functionality (auto enabled under the hood - accessible via getFunction)
- [x] handle initialStore, initialLanguage (where? existing Hook, new explicit useSSRI18n Hook, new explicit withSSRI18n HOC)
- [x] rename useReactI18n with useReactI18next here: https://github.com/i18next/react-i18next/blob/master/src/hooks/context.js#L35
- [x] create a next.js sample as proof of concept: getInitialProps as a HOC, or just a plain function provided to get that values for initialI18nStore and initialLanguage from react-i18next? (https://github.com/i18next/react-i18next/issues/671 seems a prove of working as expected)
- [x] add tests for all the new stuff
- [x] add typings for typescript

Release winter 2018/19:  https://reactjs.org/blog/2018/11/27/react-16-roadmap.html (won't be 16.7!!! https://reactjs.org/blog/2018/12/19/react-v-16-7.html)

Release seems to be 4th of February: https://github.com/facebook/react/pull/14692

schettino:
Is it going to be a full (or close to it) rewrite? Wanna give typescript a try 🚀  ? Either way, you can count on me to accomplish that!

jamuhl:
@schettino will sleep over it...

isaachinman:
I like the `context` code, and especially like `addUsedNamespaces` in `useTranslation`. This is a great, simple solution.

As far as `useSSR`, the implementation looks reasonable, but I wonder why we need to have something called `useSSR` in the first place? It seems it's just to catch the `initialLanguage`, right?

jamuhl:
yes `useSSR` is more or less a setInitialStore and setInitialLanguage (by calling i18n.changeLanguage) - i'm neither happy with the naming nor to much of exposing it...but using it in the HOC seems to have it seems reasonable -> but yes...naming it as a hook seems a little off - as it is just a function not a real hook -> so i'm open for suggestions.

isaachinman:
[...]

PR #650, opened 2018-12-27
Add hooks stub types

jleclanche:
Touches #591 

They're far from complete but better than nothing.
I had a lot of trouble with the code as there's a bunch of type issues throughout it. I would highly recommend moving the project to typescript, it'd be cleaner and easier… :)

### DIFF
--- a/hooks.d.ts
+++ b/hooks.d.ts
@@ -1,0 +1,65 @@
+import i18next from "i18next";
+import * as React from "react";
+import {Namespace, ReactI18NextOptions, TransProps} from "./index";
+
+interface HooksTransProps extends TransProps {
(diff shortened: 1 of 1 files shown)


mondayCoding:
Not sure if relevant but I tweaked index path locally in node_modules to find typings. 
Thank you for these, I was waiting for types to get to testing.

> 3 | import {Namespace, ReactI18NextOptions, TransProps} from "./index";

=>

> 3 | import {Namespace, ReactI18NextOptions, TransProps} from "./src/index";

stunaz:
@jleclanche how are we suppose to use the hooks in typescript?
when I do ` const [t, i18n] = useTranslation();`

later i want to get the current language with `i18n.language`

PR #652, opened 2018-12-30
Warn if i18next instance is a promise

adrai:
Add a warning that is triggered if the new v13 of i18next is used without being aware of the [breaking changes](https://github.com/i18next/i18next/blob/master/CHANGELOG.md#1300).
This should offer a faster feedback loop.


jamuhl:
published in react-i18next@9.0.0

PR #658, opened 2019-01-03
Fix hooks.d.ts

VincentLanglet:
The file index.d.ts have been moved


jamuhl:
published in react-i18next@9.0.2

PR #663, opened 2019-01-07
[Typescript] Fix HOC problem when picking props from WithNamespaces interface

Lukas-Kullmann:
When writing my components in, I do not want to declare all props that the `withNamespaces` HOC can pass into my components as required.

The problem is demonstrated in the `test/typescript/partial-with-namespace-pick.test.tsx` file that I added.
In this case, I want to get the `t` function from the HOC and also want to pass some other props into the component. But since I don't need (and don't want to mock in tests) properties like `i18n` and `tReady` that are provided by the HOC, I don't want to declare them on the component's props.

Before the test, the error was

```
test/typescript/partial-with-namespace-pick.test.tsx:14:33 - error TS2345: Argument of type '({ t, translationKey }: IProps) => Element' is not assignable to parameter of type 'ComponentType<WithNamespaces>'.
  Type '({ t, translationKey }: IProps) => Element' is not assignable to type 'FunctionComponent<WithNamespaces>'.
    Types of parameters '__0' and 'props' are incompatible.
      Property 'translationKey' is missing in type 'WithNamespaces & { children?: ReactNode; }' but required in type 'IProps'.

 export default withNamespaces()(MyComponent);
                                 ~~~~~~~~~~~

  test/typescript/partial-with-namespace-pick.test.tsx:6:3
       translationKey: string;
       ~~~~~~~~~~~~~~
    'translationKey' is declared here.
```

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -64,7 +64,7 @@ interface NamespaceExtractor {
 export function withNamespaces(
   namespace?: Namespace | NamespaceExtractor,
   options?: WithNamespacesOptions
-): <P extends WithNamespaces>(
+): <P extends Partial<WithNamespaces>>(
   component: React.ComponentType<P>
 ) => React.ComponentType<Subtract<P, WithNamespaces>>;
 
--- a/test/typescript/partial-with-namespace-pick.test.tsx
+++ b/test/typescript/partial-with-namespace-pick.test.tsx
@@ -1,0 +1,18 @@
+import * as React from "react";
+import { WithNamespaces, withNamespaces } from "../../src";
+
+interface IProps extends Pick<WithNamespaces, "t"> {
+  translationKey: string;
+}
+
+// translation function t can be partially picked from WithNamespaces interface
+// so that not everything needs to be passed into the component when writing tests for it
+function MyComponent({ t, translationKey }: IProps) {
+    return <div>{t(translationKey)}</div>;
+}
+
+const MyComponentWrapped = withNamespaces()(MyComponent);
+
+function App() {
+  return <MyComponentWrapped translationKey="some-translation-key" />;
+}


Lukas-Kullmann:
I guess, @schettino would be the one to review this?

So it would be nice, if you could review, merge and publish this. It's kinda blocking on our side. :)

jamuhl:
@rosskevin would love to get a ok/nok from you before merging this!

Lukas-Kullmann:
Hey @rosskevin, thanks for you feedback.
I agree that it might cause errors  when throwing around `...rest` spreads on html elements. But the thing is, that even the [react docs](https://reactjs.org/docs/jsx-in-depth.html#spread-attributes) say

> Spread attributes can be useful but they also make it easy to pass unnecessary props to components that don’t care about them or to pass invalid HTML attributes to the DOM. We recommend using this syntax sparingly.

This means to me, that you should not use rest spread if you do not exactly know what you are doing.

Also, some properties of `WithNamespaces` are optional (like `initalI18nStore` and `defaultNS`). So something like 

```typescript
interface IProps extends Pick<WithNamespaces, "t" | "tReady" | "i18n"> {
  translationKey: string;
}
```

works perfectly fine even though optional props will be passed (if they have a value).
So would you say that a change like

```diff
export interface I18nContextValues {
  i18n: i18next.i18n;
  t: i18next.TranslationFunction;
-  defaultNS?: string;
+  defaultNS: string | undefined;
-  reportNS?: string;
+  reportNS: string | undefined;
-  lng?: string;
+  lng: string | undefined;
}

export interface WithI18n extends I18nContextValues {
-  i18nOptions?: ReactI18NextOptions;
+  i18nOptions?: ReactI18NextOptions | undefined;
}

export interface WithNamespaces extends WithI18n {
  tReady: boolean;
-  initialI18nStore?: {};
+  initialI18nStore: {} | undefined;
-  initialLanguage?: string;
+  initialLanguage: string | undefined;
}
```

would be appropriate?

rosskevin:
My primary point is that you are cherry picking items to suit your pattern of use. `Partial<WithNamespaces>>` makes all keys optional, so even then you will have to perform a null check to use anything passed in (which is also likely not what you want).

Ultimately this creates a typing that is not representative of the runtime so I do not think it should be accepted.  Types are not a place to manipulate for convenience (though I understand your situation, which is why I suggest establishing your own component/pattern for your use case).

Regarding the other changes proposed above, if they are a more accurate representation of the runtime then we should use them.  If not, then we should not.

Lukas-Kullmann:
> so even then you will have to perform a null check to use anything passed in
This is not correct. I can still pick the non-optional `t` from `WithNamespaces` (as seen in the test file that I created). It's just that I do not declare everything I don't need on the props of my component. Just the `t` that I need.

rosskevin:
Yes, I missed that, but that is a minor point which isn't worth arguing if you don't address the primary point: 

Are any of the proposed types an accurate representation of the runtime? 

The answer appears to be no and therefore the PR cannot be accepted.  I may be wrong so please correct me if I am and we'll reopen this and get it right. Thanks for hashing this out with me.

PR #667, opened 2019-01-09
Update the example to respond Expo 32.0.0

shiba-hiro:
This is the PR to respond the ver up of Expo 32.

Now react-navigation v3 is available then I needed to use `createAppContainer ` with `WrappedStack`.


jamuhl:
Awesome...thank you for the update on the example - highly appreciated

PR #678, opened 2019-01-16
Fix WithNamespacesOptions `wait` requirement

### LINKED ISSUE #676: TypeScript WithNamespacesOptions.wait must be optional
**Describe the bug**
`WithNamespacesOptions.wait` is not marked as optional but docs say it has `false` as a default value

9.0.1

**To Reproduce**
```
function MyComponent() {
  return <span>Yeah!</span>
}

const translated = withNamespaces('', {})(MyComponent);
/*
[ts] Argument of type '{}' is not assignable to parameter of type 'WithNamespacesOptions'.
  Property 'wait' is missing in type '{}' but required in type 'WithNamespacesOptions'. [2345]
*/
```

**Expected behaviour**
It should be possible to set other withNamespaces options w/o explicitly setting `wait`

nikolay-borzov:
Fixes #676 

- Add test file for withNamespaces HOC

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -49,7 +49,7 @@ export interface WithNamespaces extends WithI18n {
 }
 
 export interface WithNamespacesOptions extends ReactI18NextOptions {
-  wait: boolean
+  wait?: boolean
   i18n?: i18next.i18n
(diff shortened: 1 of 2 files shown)


rosskevin:

jamuhl:
published in react-i18next@9.0.3

Issue #685, opened 2019-01-22
react-i18next throwing webpack error when initializeing using @babel/runtime 7.3.0


guiyep:
since the dependency with @babel/runtime is anything from 7.1.2 with the latest update on @babel/runtime 7.3.0 (today) the library stops working. Had to manually downgrade the version to 7.1.2 to make the library work again.

jamuhl:
? do you build react-i18next yourself? the runtime gets embedded during build time - so you should not really need/depend on it?

can you give a little more information what you are doing - and how you run into this?

jamuhl:
Just updated all dependencies -> as i guess it might be related to babel-runtime update

did a republish of react-i18next@9.0.6 does using that resolve your issue?

pedrorezende:
I guess I'm having a related problem. Just started a brand new create-react-app project yesterday, installed react-i18next and i18next and reused the same code I was using in another project. But today I'm getting the following error.

```
Uncaught TypeError: Cannot convert undefined or null to object
    at Function.getOwnPropertyDescriptors (<anonymous>)
    at _objectSpread (objectSpread.js:19)
    at setDefaults (context.js:23)
    at Object.init (context.js:37)
    at i18next.js:187
    at Array.forEach (<anonymous>)
    at I18n.init (i18next.js:186)
    at Module../src/index.js (index.js:9)
    at __webpack_require__ (bootstrap:782)
    at fn (bootstrap:150)
    at Object.0 (WarningPanel.module.scss?c11d:45)
    at __webpack_require__ (bootstrap:782)
    at checkDeferredModules (bootstrap:45)
    at Array.webpackJsonpCallback [as push] (bootstrap:32)
    at main.chunk.js:1
```

jamuhl:
could you try with react-i18next@9.0.7

guiyep:
I'm getting the exact same issue as described before. Not cloning the repo just npm installing it and then initialising it as follow:

import { reactI18nextModule } from 'react-i18next';

// i18next configuration
const config = {
  resources,
  fallbackLng: 'en-AU',
  lng: 'en-AU',
  keySeparator: false,
  interpolation: {
    format(value, format, lng) {
      return formatterExecutor(allFormatters, value, format, lng);
    },
    escapeValue: false,
  },
};
const i18next = i18n.use(reactI18nextModule).init(config);

jamuhl:
[...]

PR #686, opened 2019-01-22
Fix runtime test while no react options.

jingfei:
Fix runtime test while no react i18next special options in init.

### DIFF
--- a/src/context.js
+++ b/src/context.js
@@ -35,7 +35,7 @@ export const reactI18nextModule = {
   type: '3rdParty',
 
   init(instance) {
-    setDefaults(instance.options.react);
+    setDefaults(instance.options.react || {});
     setI18n(instance);
   },
 };


jamuhl:
Wouldn't it be cleaner to change https://github.com/i18next/react-i18next/blob/master/src/context.js#L18 to

```js
export function setDefaults(options = {}) {
```

jingfei:
I added there because `options.react` is optional. I think it depends on the design of `setDefaults`.

jamuhl:
beside even https://github.com/i18next/react-i18next/blob/master/src/context.js#L19

```js
defaultOptions = { ...defaultOptions, ...options };
```

Does not lead to an exception when `options` is undefined 

when do you run into an issue with this?

jingfei:
> when do you run into an issue with this?
When there's no `react` option in `init`, and get the exception since `options` is undefined. 

My opinion is maybe there exists settings which are not optional. Thus exception cannot be ignored.

jamuhl:
options can't be undefined

external modules get called here: https://github.com/i18next/i18next/blob/master/src/i18next.js#L92

after setting this options: https://github.com/i18next/i18next/blob/master/src/i18next.js#L40

so `instance.options` is defined and even if `instance.options.react` is undefined it can't really lead to an exception

do you call this different?!? seems really more a constructed edge case while doing something "wrong"

jamuhl:
at least i do not get an exception here on the sample: https://github.com/i18next/react-i18next/tree/master/example/react

when removing react options here: https://github.com/i18next/react-i18next/blob/master/example/react/src/i18n.js#L28

So would be great to get a sample to reproduce this is needed.

jingfei:
Sorry it was typo, `inatance.options.react` is undefined, and get exceptions while doing `...options` in https://github.com/i18next/react-i18next/blob/master/src/context.js#L19.

FYI: Exception log
```
react-i18next/node_modules/@babel/runtime/helpers/objectSpread.js:19
      Object.defineProperties(target, Object.getOwnPropertyDescriptors(arguments[i]));
```

jamuhl:
Just updated all dependencies -> as i guess it might be related to babel-runtime update

did a republish of react-i18next@9.0.6 does using that resolve your issue?

jamuhl:
seems related: https://github.com/i18next/react-i18next/issues/685

jamuhl:
should be fixed by https://github.com/i18next/react-i18next/commit/bf830d965b26ccf75afaaf6e4741cebf48d74827

in react-i18next@9.0.7 (still unclear where and why this happens...can't reproduce this in my samples)

jingfei:
It is solved after version updated. Thank you. :)

PR #691, opened 2019-01-23
TS generics fixes related to i18next/1180

rosskevin:
Depends on https://github.com/i18next/i18next/pull/1180


rosskevin:
No release for this is necessary, just test update to stay current with i18next ts changes. 

/cc @tkow

PR #698, opened 2019-01-25
fix: useTranslation hooks typing

panjiesw:
This update useTranslation hook return type to use `TFunction`


jamuhl:
will publish this with the next patch? or do you need this now?

panjiesw:
Would be nice to have this soon I guess?

jamuhl:
published in react-i18next@9.0.9

PR #500, opened 2018-08-23
Added reportNS function to I18NextProvider to report used namespaces

olivierpascal:
Related to #499

PR #699, opened 2019-01-27
Add TS definition for reportNS prop

johnnynia:
The reportNS prop of I18nextProvider is missing in I18nextProviderProps.

https://github.com/i18next/react-i18next/blob/5e929b56709d92c1bce87cb939bb00800ebfafb0/src/I18nextProvider.js#L14

It was added in PR https://github.com/i18next/react-i18next/pull/500


rosskevin:

jamuhl:
published in react-i18next@9.0.10

Issue #701, opened 2019-01-28
allow 'TemplateStringsArray' as parameter of t function in type declaration


drewxiuxiuxiu:
**Is your feature request related to a problem? Please describe.**
I'd like to reduce a pair of parenthesis by using string literal: 

```
t`some_key` // instead of t('some_key')
```
which is 98% of the case in my code base. It doesn't effect anything, only need the type support. 

I can raise a PR if it is acceptable.

**Describe the solution you'd like**
simply add types to the declaration file

**Describe alternatives you've considered**

**Additional context**

jamuhl:
for me a PR would be ok? @rosskevin opinions on this?

rosskevin:
I'd like to see the PR - definitely not against it if it can be supported in a overloaded way.  @drewxiuxiuxiu you'll have to modify the `WithT` and `TFunction`, then add a test to `t.external.test.ts` to attempt both variations.

jamuhl:
please start adding changes related to the v9 version to the v9.x.x branch (master will be now for the upcoming v10 hooks)

--> https://github.com/i18next/react-i18next/tree/v9.x.x

noru:
#705 PR submitted.

It actually makes more sense to update the types on `i18next` side, and it would be easier. Will check on the other side.

rosskevin:
Closing in favor of an issue in i18next.  When I originally commented, I made a mistake on mobile and assumed this was in that repo.

PR #705, opened 2019-01-30
#701, allow T function to accept TemplateStringsArray

noru:


rosskevin:
This PR belongs in i18next

Issue #712, opened 2019-02-06
Design decision issue: useTranslation hook returning array vs object


kachkaev:
👋  I've just discovered v10 (the hooks one) and am pretty excited by it! Here's a question about one architectural decision that has come to my mind: _Why is the result of useTranslation() an array rather than an object?_

```js
const [ t, i18n ] = useTranslation();
// vs.
const { t, i18n } = useTranslation();
```
 
Arrays are good for things that can have arbitrary names, e.g. `[x, setX] = useState()`, which is not the case for `react-i18next`'s `useTranslation()`. As a developer, I'd like `t` to be `t` in all components, same for `i18n`. Doing something like `const [ abc, def ] = useTranslation();` would mean the component's code is harder to understand by the developers, which a sign of not following a good practice. On the other hand, if the result of the hook is an object, I can do  `const { i18n } = useTranslation()`, e.g. when I only need to know the current language in a component. No need to have `[, i18n]` in this case.

Alternatively, the hook could be like

```js
const i18n = useI18n();

// when only t is needed
const { t } = useI18n();

// i18n + shortcut to t
const i18n = useI18n();
const t = i18n.t;
```

@jamuhl  If I've missed a conversation about the hook design, could please point me to one? Otherwise, what are your thoughts on the current hook design? How did you justify it?

jamuhl:
that architectural decision was made by react: https://reactjs.org/docs/hooks-intro.html

kachkaev:
Hooks don't have to return an array – even [built-in](https://reactjs.org/docs/hooks-reference.html) `useContext()`, `useMemo()`, `useCallback()` etc. do not do that.

You can check out https://usehooks.com/ with a few examples, some of which returning arrays, some – objects. E.g. for [`useHistory()`](https://usehooks.com/useHistory/) one it makes sense to return a spreadable object rather than an array – otherwise there are too many items to cycle through before you get to the one you need. Looking at various hooks, I can say that there are more hooks that don't return an array than the ones that do. This is what makes me wonder what return typ a hook in `react-i18next` should have. 🤔

jamuhl:
`useState` and `useReducer` do -> so i guess it will be what users will expect

kachkaev:
Although these two hooks are common, I expect there will soon be a whole swarm of other commonly hooks in the npm ecosystem, which won't be using this return pattern (for a good reason). So I don't think there will be an expectation issue if a hook from `react-i18next` returns something other than an array.

[...]

PR #714, opened 2019-02-06
Added possibility to use useTranlation returning Array or Object

adrai:
Added possibility to use useTranlation not only expecting an array but also an object. Should help #712

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -42,10 +42,16 @@ export function useSSR(initialI18nStore: any, initialLanguage: any): void;
 export interface UseTranslationOptions {
   i18n?: i18next.i18n;
 }
+export interface UseTranslationResponse extends Array<i18next.TFunction | i18next.i18n> {
+  0: i18next.TFunction;
+  1: i18next.i18n;
+  t: i18next.TFunction;
+  i18n: i18next.i18n;
+}
 export function useTranslation(
   ns?: Namespace,
   options?: UseTranslationOptions,
-): [i18next.TFunction, i18next.i18n];
+): UseTranslationResponse;
 
 // Need to see usage to improve this
 export function withSSR(): (
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -42,12 +42,10 @@ export function useSSR(initialI18nStore: any, initialLanguage: any): void;
 export interface UseTranslationOptions {
   i18n?: i18next.i18n;
 }
-export interface UseTranslationResponse extends Array<i18next.TFunction | i18next.i18n> {
-  0: i18next.TFunction;
-  1: i18next.i18n;
+export type UseTranslationResponse = [i18next.TFunction, i18next.i18n] & {
   t: i18next.TFunction;
   i18n: i18next.i18n;
-}
+};
 export function useTranslation(
   ns?: Namespace,
   options?: UseTranslationOptions,
--- a/test/typescript/examples.test.tsx
+++ b/test/typescript/examples.test.tsx
@@ -22,6 +22,7 @@ function MyComponent() {
 // page uses the hook
 function Page() {
   const [t, i18n] = useTranslation();
+  // const { t, i18n } = useTranslation();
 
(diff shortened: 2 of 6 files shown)


src/index.d.ts
```diff
@@ -42,10 +42,16 @@ export function useSSR(initialI18nStore: any, initialLanguage: any): void;
 export interface UseTranslationOptions {
   i18n?: i18next.i18n;
 }
+export interface UseTranslationResponse extends Array<i18next.TFunction | i18next.i18n> {
```
adrai: @rosskevin @schettino I've no clue about typescript... Is this ok?

jamuhl:
@kachkaev opinions about letting users choose? we could provide both options

kachkaev:
What if there is `const i18n = useI18n()?` as an alternative? It can be used as const { t } = useI18n()` if `t` is bound to `i18n` correctly (no issues with `this`)? The original hook can be marked as deprecated over time if it’s perceived as a less convenient one.

jamuhl:
@kachkaev i want to avoid another hook. As 95% of the times you like to get some t function based on a given namespace i think this is prefered

```js
const { t } = useTranslation('ns1');
const [t] = useTranslation('ns1');
```

The only case you like to get the i18n instance is where you need current language or like to access the changeLanguage function. Chances are high you need t function in this component too.

So your approach would be:

```js
const i18n = useI18n();
const { t } = i18n;
```

while you could just:

```js
const { t, i18n } = useTranslation();
```

Beside that what we do for HOC and render prop? withI18n?!?

```js
function Extended({ i18n }) {
   const { t } = i18n; 
}
```

src/index.d.ts
kachkaev: @adrai you can try:

```ts
type UseTranslationResponse = [i18next.TFunction, i18next.i18n] & {
  t: i18next.TFunction;
  i18n: i18next.i18n;
}
```

kachkaev:
Oh, I did not realise that JavaScript would allow us to return something that's both an array and an object! In this case, keeping just `useTranslation` makes perfect sense!

src/index.d.ts
adrai: thx, just updated it

adrai:
Yes, JavaScript allows to extend arrays.
With this PR you can now choose to do:
`const [ t, i18n ] = useTranslation();`
or
`const { t, i18n } = useTranslation();`

kachkaev:
@jamuhl WDYT of updating docs too and encourage users to stick with `{ t, i18n }`? The array can still be still supported for BC, but this will be a 'hidden feature'.

test/typescript/examples.test.tsx
```diff
@@ -22,6 +22,7 @@ function MyComponent() {
 // page uses the hook
 function Page() {
   const [t, i18n] = useTranslation();
+  // const { t, i18n } = useTranslation();
```
rosskevin: please remove comment

rosskevin: Actually, we need both if you want to merge this PR.  If you are expecting it to return both.

rosskevin: I'll push some test changes.

rosskevin:
> @jamuhl WDYT of updating docs too and encourage users to stick with `{ t, i18n }`? The array can still be still supported for BC, but this will be a 'hidden feature'.

I have added explicit TS tests for both usages, and I agree that the object usage should be promoted as the default (my commit reflects that) especially for js users not using types.

src/index.d.ts
rosskevin: This is bizarre - but I confirmed it typechecks correctly.

jamuhl:
@rosskevin @kachkaev @adrai yes...i will update the docs / samples in favor to using object spread `const { t, i18n } = useTranslation();`

jamuhl:
published in react-i18next@10.0.1

jamuhl:
https://react.i18next.com/latest/usetranslation-hook and all the other places in the docs should reflect this change

Issue #718, opened 2019-02-07
v10 I18nextProvider typings missing


panjiesw:
**Describe the bug**
TypeScript typings doesn't include `I18nextProvider`.

**Occurs in react-i18next version**
v10.0.0

**To Reproduce**

1. `import {I18nextProvider} from 'react-i18next` in TypeScript
2. See error

```
Module '"node_modules/react-i18next/src"' has no exported member 'I18nextProvider'.ts(2305)
```

**Expected behaviour**
Typings should include `I18nextProvider`

adrai:
Can you suggest a PR? And perhaps @rosskevin or @schettino could review...

panjiesw:
Not now unfortunately, maybe later this night (UTC+7)

jkeruzec:
Could you provide an upgrade guide from V9 to V10 please ? 

I saw that reactI18nextModule became initReactI18next.

I lost my withNamespaces and I don't what is replacing this ?

Brisstor:
@jkeruzec seems like withTranslation is replacement for withNamespaces.
However, I didn't see any warnings about v10 containing some breaking changes...
I thing it would be nice to have some notification about that :)
Anyway, thanks for fast update!

jamuhl:
> I didn't see any warnings about v10 containing some breaking changes

@Brisstor semver?!? 9.x.x -> 10.0.0 = Breaking API else it would be just 9.1.0 (additional hooks no breaking)

Brisstor:
> > I didn't see any warnings about v10 containing some breaking changes
> 
> @Brisstor semver?!? 9.x.x -> 10.0.0 = Breaking API else it would be just 9.1.0 (additional hooks no breaking)
You're right, that makes sense, thanks

jamuhl:
@jkeruzec @Brisstor https://react.i18next.com/latest/migrating-v9-to-v10 hopefully helps a little - contribution with your migration experiences would be very welcome.

danielsogl:
I created an PR to fix that issue. I had the same problem after migrating to V10.

PR #721, opened 2019-02-07
Add missing I18nextProvider interface to typings

danielsogl:
closes: #718

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -76,7 +76,7 @@ export interface I18nextProviderProps {
   i18n: i18next.i18n;
 }
 
-export const I18nextProvider: React.ComponentClass<I18nextProviderProps>;
+export const I18nextProvider: React.FunctionComponent<I18nextProviderProps>;
 
 export interface TranslationProps {
   children: (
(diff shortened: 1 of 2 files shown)


jamuhl:
v10 only has i18n as prop!

danielsogl:
Correct! I updated the PR

src/index.d.ts
```diff
@@ -72,6 +72,12 @@ export function withTranslation(
   component: React.ComponentType<P>,
 ) => React.ComponentType<Omit<P, keyof WithTranslation>>;
 
+export interface I18nextProviderProps {
+  i18n: i18next.i18n;
+}
+
+export const I18nextProvider: React.ComponentClass<I18nextProviderProps>;
```
panjiesw: Shouldn't this be `React.FunctionComponent<I18nextProviderProps>` or `React.FC<I18nextProviderProps>`? Because basically the component is a function component, not a class one.

danielsogl: Then it should be React.FunctionComponent

danielsogl:
@rosskevin I added a basic test example.

rosskevin:
Thanks @danielsogl 

danielsogl:
That would be great!

jamuhl:
published in react-i18next@10.0.1

Issue #727, opened 2019-02-09
Optional children for Trans component


maksis:
https://react.i18next.com/latest/trans-component#alternative-usage fails with error `Property 'children' is missing in type '{ defaults: string; values: { what: string; }; components: Element[]; }' but required in type 'TransProps'.` when using version 10.0.3

basslagter:
Yes, same here. Fixed it by providing a child <Trans>{}</Trans> but does not seem like a sustainable solution

jamuhl:
A pull request would be welcome...guess @rosskevin will then review that

menthena:
Could you please check this? https://github.com/i18next/react-i18next/pull/728

basslagter:
Thanks!

PR #728, opened 2019-02-11
Make children optional within TransProps

menthena:
Fixes https://github.com/i18next/react-i18next/issues/727


rosskevin:
@menthena please add another usage example that covers the case in `test/typescript`

rosskevin:

jamuhl:
published in react-i18next@10.0.5

Issue #733, opened 2019-02-13
Translations loaded despite using initialStore prop


crash7:
Hello! I was updating the razzle-ssr example to use Hooks and noticed that even with the store pre-populated with the initial data, the translations are still trying to be loaded (triggering a suspense, making the page render again discarding the SSR render). Is this expected? Or a bug in xhr-backend?

I have another project with v9 and see the same behavior there.

The WIP can be found here: https://github.com/crash7/react-i18next/tree/example/razzle-ssr/example/razzle-ssr

jamuhl:
highly appreciate your effort getting that sample up again.

https://github.com/crash7/react-i18next/blob/example/razzle-ssr/example/razzle-ssr/src/server.js#L31

you're only telling i18next to preload languages...but it has no information about which namespaces it has to load, init with:

```js
i18next.init({
  ns: ['ns1', 'ns2'], // all namespaces used
  defaultNS: 'ns1', // the default to use
  //...
```

I once had a helper in one of the next.js samples:
https://github.com/i18next/react-i18next/blob/v8.3.3/example/nextjs/lib/getAllNamespaces.js

used like:
https://github.com/i18next/react-i18next/blob/v8.3.3/example/nextjs/server.js#L21

might be helpful?

crash7:
Actually the full i18n config is here: https://github.com/crash7/react-i18next/blob/example/razzle-ssr/example/razzle-ssr/src/i18n.js#L11

And this is the output of the debug info in the server console: 
```
i18next: initialized { debug: true,
  initImmediate: false,
  ns: [ 'translations' ],
  defaultNS: 'translations',
```

I didn't know you could extend the config doing that, but even if I declare the ns in the server config, the behavior is the same.

```diff
  .init(
    {
      preload: ['en', 'de'],
+     ns: ['translations'],
+     defaultNS: 'translations',
      backend: {
        loadPath: `${appSrc}/locales/{{lng}}/{{ns}}.json`,
        addPath: `${appSrc}/locales/{{lng}}/{{ns}}.missing.json`,
      },
    },
    () => {
      server
```

Another thing that validates the config, is that I get the translations in SSR (this means that the i18n is getting preloaded as expected) and also the initialStore data in the source HTML:

```
[...]

PR #740, opened 2019-02-16
docs: add razzle-ssr example with hooks

crash7:
Rewrite the old v9 razzle-ssr example with the v10 changes and hooks.

Thank you @jamuhl for all the help debugging the Suspense problem!

closes https://github.com/i18next/react-i18next/issues/733


jamuhl:
Awesome...thanks a lot for bringing this back to v10

PR #745, opened 2019-02-19
Add React Native example using hooks 

pontusab:
Here is a basic example for React Native that uses v10 and hooks.


jamuhl:
you're awesome...thanks a lot for sharing

PR #753, opened 2019-02-21
Add ready/tReady type definitions

### LINKED ISSUE #751: TypeScript type definitions hook/HOC aren't correct
As the title already states the TypeScript type definitions aren't correct for both the hook and HOC, mainly the return type and prop definitions.

In both cases the types are missing the `ready` and `tReady` typings which I believe is just a boolean?

I read in some other issues that to fix these type definitions I should create a pull request to DefinitelyTyped but since the type definitions are now maintained here I think it should be fixed here.

Unless there is something I'm overlooking the fix is as simple as doing the following:
```typescript
export type UseTranslationResponse = [i18next.TFunction, i18next.i18n, boolean] & {
  t: i18next.TFunction;
  i18n: i18next.i18n;
  ready: boolean;
};

export interface WithTranslation extends i18next.WithT {
  i18n: i18next.i18n;
  tReady: boolean;
}
```

traverse:
Adds missing type definitions for the `ready` return value of the `useTranslation` hook and the `tReady` prop added by the `withTranslation` HOC, fixes #751.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "babel-plugin-macros": "^2.5.0",
     "babel-plugin-tester": "^5.5.2",
     "coveralls": "^3.0.2",
-    "dtslint": "^0.5.3",
+    "dtslint": "0.5.3",
     "enzyme": "^3.8.0",
     "enzyme-adapter-react-16": "^1.9.1",
     "enzyme-adapter-react-helper": "^1.3.2",
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "babel-plugin-macros": "^2.5.0",
     "babel-plugin-tester": "^5.5.2",
     "coveralls": "^3.0.2",
-    "dtslint": "^0.4.2",
+    "dtslint": "^0.5.3",
     "enzyme": "^3.8.0",
     "enzyme-adapter-react-16": "^1.9.1",
     "enzyme-adapter-react-helper": "^1.3.2",
@@ -74,7 +74,7 @@
     "rollup-plugin-replace": "^2.1.0",
     "rollup-plugin-terser": "^4.0.4",
     "sinon": "^7.2.3",
-    "tslint": "^5.12.1",
+    "tslint": "^5.13.1",
     "typescript": "^3.3.3",
     "yargs": "12.0.5"
   },
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
     "rollup-plugin-replace": "^2.1.0",
     "rollup-plugin-terser": "^4.0.4",
     "sinon": "^7.2.3",
-    "tslint": "^5.13.1",
+    "tslint": "5.13.1",
     "typescript": "^3.3.3",
     "yargs": "12.0.5"
   },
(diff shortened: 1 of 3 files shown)


traverse:
@rosskevin do you mean something like the following in `examples.test.tsx`:
```typescript
// Component that uses ready to wait until translations are loaded because useSuspense is set to false
function ComponentNotUsingSuspense() {
  const { t, ready } = useTranslation();
  return (
    <div>
      {ready && t('key1')}
    </div>
  );
}

// Class based component that uses tReady to wait until translations are loaded because useSuspense is set to false
class ClassComponentNotUsingSuspense extends React.Component<WithTranslation> {
  render() {
    const { t, tReady } = this.props;
    return <h2>{tReady && t('title')}</h2>;
  }
}
```

rosskevin:
@traverse you can add both of those, but be sure to exercise them (just add them inside a closure):

```tsx
function suspenseUsage() {
    // components code here

   // exercise components code
   return (
        <div>
            <ComponentNotUsingSuspense />
            <ClassComponentNotUsingSuspense />
         </div>
    )
}
```

traverse:
Ok will do, thanks!

jamuhl:
@traverse any update on this?

traverse:
@jamuhl sorry about not giving an update on this, I was quite pressed for time last week and was bedridden over the weekend 🤒. I'm going to be spending some time on this today.

traverse:
@rosskevin any idea why the CI build might be failing? The tests are running fine locally 🤔

rosskevin:
@traverse - it seems bizarre. I have tracked down dtslint to be relying on `TypeScript@next`, which might reveal the error on ci and not local.  We are currently testing with the release of `3.3.3`.

I'm looking into why dtslint is on `next` - because I believe that is a nightly build.

rosskevin:
Issue https://github.com/Microsoft/dtslint/issues/159

rosskevin:
@traverse this all looks good locally, I agree.  I tried several things, all with full rebuilds, all which succeed locally but only fail on travis.  

 @jamuh I don't have much confidence in travis at this point.  I checked and there are no caches configured so I'm not sure what else to do there. My assumption at this point is that any change in this repo (even non-ts) is going to fail on travis due to some environmental situation there.

I'm happy to merge this without travis success but worried that we will see more of the same.  Thoughts?

traverse:
@rosskevin I'm not sure if it would work in this case and if there's something similar for NPM but with Yarn you can use [resolutions](https://yarnpkg.com/lang/en/docs/selective-version-resolutions/) to overwrite specific dependency versions.

rosskevin:
@traverse I tested with typescript next locally and it works, so it isn't a version conflict. Something on Travis.

traverse:
@rosskevin It's also quite weird it's failing to find both `withTranslation` and `WithTranslation` but is finding `useTranslation` I'd assume it would either find none or all.

jamuhl:
works on https://circleci.com/gh/i18next/react-i18next/3

currently switching over to circleci

jamuhl:
Failed local at my place too: https://github.com/i18next/react-i18next/commit/ef1ee56e4d9114dd477ca7f270906ba7547121b7 fixes it

jamuhl:
https://circleci.com/gh/i18next/react-i18next/9

jamuhl:
published in react-i18next@10.3.0

PR #756, opened 2019-02-25
Update index.d.ts

hchood:
I believe this change should have been made here: https://github.com/hchood/react-i18next/commit/c2d4187f0f2b53373267cd933405bb65d7e05a4b#diff-b739f95ace6bbf2f0fd5e8cf1dab97d8R61

I'm not totally sure, however, since there's also this: https://github.com/i18next/react-i18next/blob/master/react-i18next.js#L778-L796 which is I think the one that is actually called.

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -37,7 +37,7 @@ export interface TransProps extends Partial<i18next.WithT> {
 }
 export function Trans(props: TransProps): any;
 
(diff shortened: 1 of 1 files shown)


hchood:
Sorry, made this accidentally!

jamuhl:
Still a good chance to say:

PR #774, opened 2019-03-08
Resolve #773 by sending console as the first argument to apply

### LINKED ISSUE #773: Broken in IE11 due to calling console.warn.apply with null
**Describe the bug**
When transpiling to IE11, a console.warn statement is being called with the specific footprint of `console.warn.apply(null, args)` in the function warn

```js
  function warn() {
    if (console && console.warn) {
      for (var _len = arguments.length, args = new Array(_len), _key = 0; _key < _len; _key++) {
        args[_key] = arguments[_key];
      }

      if (typeof args[0] === 'string') args[0] = `react-i18next:: ${args[0]}`;
      console.warn.apply(null, args);
    }
  }
```

from our package.json: "react-i18next": "^10.2.0"

**To Reproduce**

1. attempt to cause a warning in IE11

**Expected behaviour**
The script continues to execute

**Actual Behavior**
The script on IE11 fails due to incorrectly calling console.warn with a null context

**Screenshots**
screen shot 2019-03-07 at 11 48 20 am

- Device: Windows 7
- Browser: IE11

**Additional context**
This cannot be recreated when dev tools are open due to the way dev tools wraps the console methods when injecting itself.

We have not checked for other console statement in the package

Any console statements called with `apply` with a context of `null` will fail.

This should be a simple fix of just replacing `null` in these `apply` statements with `console`

skatcat31:
Resolves #773 

The bug was due to sending `null` as the context to `console.warn.apply`.

Fixed by passing `console` as the context.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -1,7 +1,7 @@
 export function warn(...args) {
   if (console && console.warn) {
     if (typeof args[0] === 'string') args[0] = `react-i18next:: ${args[0]}`;
-    console.warn.apply(null, args);
+    console.warn.apply(console, args);
   }
 }
 


skatcat31:
Apparently code climate thinks a spread operator would be better. That may be true, but I merely wanted to fix a bug my team ran into with this package without changing the code base to heavily.

jamuhl:
thank you for taking the time to create the PR. Was merged and published to npm with react-i18next@10.3.1

PR #798, opened 2019-03-22
Fix the displayName of HOC components

SimeonC:
Updating the HOC components to have displayNames consistent with the react docs definition: https://reactjs.org/docs/higher-order-components.html#convention-wrap-the-display-name-for-easy-debugging
This affects some dev tools like react-cosmos that use this format to detect what the "Original" component name is. Also standard is good.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -69,3 +69,11 @@ export function hasLoadedNamespace(ns, i18n) {
 
   return false;
 }
+
+export function getDisplayName(Component) {
+  return (
+    Component.displayName ||
+    Component.name ||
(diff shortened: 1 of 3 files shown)


jamuhl:
I'm not a big fan of adding tons of dependencies...in this case the function is very small -> so please add the function to utils and remove the npm dependency

SimeonC:
Fair enough, have updated removing the dependency.
I guess this is small enough having multiple versions of the same code in your JS Bundle won't matter so much.

jamuhl:
published in react-i18next@10.5.3

PR #818, opened 2019-04-05
Fix Trans component ignore default options

yashigani:
react-i18next has some default options but Trans component ignore it.
Reproduce it project: https://codesandbox.io/s/1zxox032q

### DIFF
--- a/src/Trans.js
+++ b/src/Trans.js
@@ -1,6 +1,6 @@
 import React, { useContext } from 'react';
 import HTML from 'html-parse-stringify2';
-import { getI18n, getHasUsedI18nextProvider, I18nContext } from './context';
+import { getI18n, getHasUsedI18nextProvider, I18nContext, getDefaults } from './context';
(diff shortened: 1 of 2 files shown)


jamuhl:
Hm...now you ignore the passed in options

if need defaults + passed in it should be:

```js
init(instance) {
    setDefaults({ ...getDefaults(), ...instance.options.react});	    
    setI18n(instance);
  },
};
```

yashigani:
I seem `setDefaults` method merge options into `defaultOptions`.

```
export function setDefaults(options = {}) {
  defaultOptions = { ...defaultOptions, ...options };
}
```

jamuhl:
right...i see now what you mean...the Trans component also takes stuff from i18nextOptions directly...for cases passed in i18n via props but using "setDefaults" does not change your options...

guess will merge options in Trans...please keep this open...will provide a fix asap

jamuhl:
should be fixed in react-i18next@10.6.2

Issue #823, opened 2019-04-11
useSSR doesn't change language server side


Swiip:
**Describe the bug**
Using `useSSR` with Next. I cache user language selection with a cookie. The server always use the same language server side which leads to hydration problems.

Linked to #671 and probably #817 

Looking at the code of `useSSR`, it seems that the use of `i18n.initializedLanguageOnce` prevent to change language. It's ok client side but server side, the same server with the same instance of i18next will be reused to handle requests asking with different languages.

**Occurs in react-i18next version**
10.1.1

**To Reproduce**
1. Project with Next.js
2. Use of react-i18next and useSSR
3. Use cookie strategy to store user prefered language
4. Render any page in a first language
5. Change language
6. Refresh

**Expected behaviour**
`useSSR` server side should be able to respond with the detected language, not keeping the first language used

**OS (please complete the following information):**

- Device: MacBook Pro 15'
- Browser: Chrome 73

**Additional context**
-

jamuhl:
Do you use the same instance of i18next on serverside render do you not use the i18n instance that you get from req.i18n using the express middleware?

Swiip:
I use my own instance indeed. I didn't thought the middleware will give me a new one each time. To have the full context, here is the gist of my Next / i18next HoC.

https://gist.github.com/Swiip/457f50cb4c4da0698f39cc06db2cad9e

jamuhl:
using that is crucial -> as think of two requests doing async operations before doing send the response both use your one singleton i18next instance and change the language on it...you will get wrong responses.

for react-i18next + next.js i highly recommend: https://github.com/isaachinman/next-i18next

jamuhl:
exposing https://github.com/i18next/react-i18next/blob/master/src/context.js#L17 here https://github.com/i18next/react-i18next/blob/master/src/index.js#L9 would be enough....

Would you like to provide a PR?

Swiip:
Yep, exactly, I will send the PR

PR #829, opened 2019-04-17
export I18nextContext to allow users to get the context instance of i18next

Swiip:
Linked to #823, it could be useful for users to have access to the i18next instance in context.


jamuhl:
was published in react-i18next@10.8.1

PR #852, opened 2019-05-16
Fix useEffect mount/unmount usage

jigsawye:
The second params of useEffect should be `[]`, otherwise it will be called when re-render every time.


jamuhl:
You're right...will be merged and published asap. Thanks for catching that one...

jamuhl:
was published in react-i18next@10.9.1

PR #853, opened 2019-05-17
HOC: expose wrapped component as WrappedComponent property

naholyr:
Hi, this PR will add a public static property `WrappedComponent` (tests included) to access original component.

It looks like a *de facto* standard in *higher order components*, at least that's what `react-router` and `react-redux` do, so I expected any higher order component to do the same. This allows to access original component in a standardized way and access its original static properties (in my case, a `loadData` property used for server-side data fetching, but that could be anything else). Without this property, I can't use your HOC with mine as it "breaks the chain". This PR includes implementation and test :+1:

Thanks for your work!

Note: this is quite the same PR which was merged last week in dunglas/react-esi#6

### DIFF
--- a/src/withSSR.js
+++ b/src/withSSR.js
@@ -15,6 +15,7 @@ export function withSSR() {
 
     I18nextWithSSR.getInitialProps = composeInitialProps(WrappedComponent);
     I18nextWithSSR.displayName = `withI18nextSSR(${getDisplayName(WrappedComponent)})`;
+    I18nextWithSSR.WrappedComponent = WrappedComponent;
 
     return I18nextWithSSR;
   };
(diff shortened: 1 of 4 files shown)


jamuhl:
For consistency in the code - can you do the same for: https://github.com/i18next/react-i18next/blob/master/src/withSSR.js

naholyr:
Obviously, I'll do it tomorrow ;)
I did not give attention to this one as he's not as used, but you're right.

Le sam. 18 mai 2019 à 14:30, Jan Mühlemann <notifications@github.com> a
écrit :

naholyr:
PR updated with `withSSR` (force-push to keep it single-commit)

jamuhl:
thank you...was published in react-i18next@10.10.0

PR #859, opened 2019-05-28
wip: feat(Trans): allow components prop to be an object

### LINKED ISSUE #833: Support passing objects to <Trans components={{ ... }} />
Hi there! First of all, thank you for all the work you put into `i18next`! It is by far the best i18n lib we've come across, and it just keeps getting better and better :)

**Is your feature request related to a problem? Please describe.**

I'm running into some issues with the numeric (`<0>`) placeholders in the `<Trans components={[...]} />` prop. 
First of all, the translation tool we use has first-class support for html placeholders, meaning that during the translation process they automatically get detected and copied to the target translation. Without a mechanism like this you very quickly run into awkward scenarios with for example RTL input. However, this detection is built on top of an X(HT)ML parser, and the XML spec dictates that tags are only valid if they start with a letter. As a result the numeric `<0>` placeholders are not picked up and the detection does not work in this case.

Secondly, I'm noticing that numeric placeholders sometimes convey too little context to translators. Given the following 2 strings:

* `<0>{{firstName}} {{lastName}}</0> added a new <1>comment</1> to article <2>{{articleTitle}}</2>`
* `<profileLink>{{firstName}} {{lastName}}</profileLink> added a new <commentLink>comment</commentLink> to article <articleLink>{{articleTitle}}</articleLink>`

I personally prefer the latter one. It's a bit more verbose, but verbosity is important when involving external translators into a project as they need additional context to figure out what is what. In addition, it looks more familiar to html which they generally already have some notion of, + it's a bit more resilient to change as it doesn't rely on order.

**Describe the solution you'd like**

I was wondering if you'd be open to support a secondary object syntax for the `components` prop, i.e.:

```js
<Trans
  i18nKey="myKey"
  components={{
    articleLink: <a href="..." />,
    commentlink: <a href="..." />,
    profileLink: <a href="..." />
  }}
  values={values}
/>
```

Which would enable the use of the "named placeholder" method above. I can have a stab at this myself, just will have to dig into the `Trans` component a bit first :)

iamandrewluca:
* [ ] allow `Trans.props.components` to be an object
* [x] all previous tests pass
* [ ] write tests
* [ ] add examples/docs

Notes:
* do not use [void-elements](http://w3c.github.io/html/syntax.html#void-elements) as object key because [html-parse-stringify2](https://github.com/rayd/html-parse-stringify2#1-tag) will parse them as empty

Blocker rayd/html-parse-stringify2/pull/20
Closes #833

### DIFF
--- a/src/Trans.js
+++ b/src/Trans.js
@@ -1,5 +1,5 @@
-import React, { useContext } from 'react';
-import HTML from 'html-parse-stringify2';
+import { useContext, isValidElement, createElement, cloneElement } from 'react';
+import { parse } from 'html-parse-stringify2';
 import { getI18n, I18nContext, getDefaults } from './context';
 import { warn, warnOnce } from './utils';
 
@@ -14,9 +14,10 @@ function getChildren(node) {
 
 function hasValidReactChildren(children) {
   if (Object.prototype.toString.call(children) !== '[object Array]') return false;
-  return children.every(child => React.isValidElement(child));
+  return children.every(child => isValidElement(child));
 }
 
+// React.Children.toArray cannot replace this, will fail at a child of object type
 function getAsArray(data) {
   return Array.isArray(data) ? data : [data];
 }
@@ -24,18 +25,15 @@ function getAsArray(data) {
 export function nodesToString(children, i18nOptions) {
   if (!children) return '';
   let stringNode = '';
-
-  // do not use `React.Children.toArray`, will fail at object children
-  const childrenArray = getAsArray(children);
   const keepArray = i18nOptions.transKeepBasicHtmlNodesFor || [];
 
   // e.g. lorem <br/> ipsum {{ messageCount, format }} dolor <strong>bold</strong> amet
-  childrenArray.forEach((child, childIndex) => {
+  getAsArray(children).forEach((child, childIndex) => {
     if (typeof child === 'string') {
       // actual e.g. lorem
       // expected e.g. lorem
       stringNode += `${child}`;
-    } else if (React.isValidElement(child)) {
+    } else if (isValidElement(child)) {
       const childPropsCount = Object.keys(child.props).length;
       const shouldKeepChild = keepArray.indexOf(child.type) > -1;
       const childChildren = child.props.children;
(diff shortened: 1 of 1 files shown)


jamuhl:
ping me when tests were added

pgom:
@iamandrewluca what's the status on this? Would love to get this merged. If you don't have availability, I can jump in and update what's necessary.

iamandrewluca:
@pgom I can finish this, but object for now should not have keys from void elements list

https://html.spec.whatwg.org/multipage/syntax.html#void-elements

pgom:
@iamandrewluca Got it. Can you give any ETA? Do you need some help?

pgom:
Hey @iamandrewluca, just checking on this. Can you still move forward with it?

iamandrewluca:
@pgom I want, but I don't have the time :slightly_smiling_face: Also files my implementation was based on, are changed now, having conflicts, and have to start again from start.

pgom:
Ok, let me know if you need some help 😀

iamandrewluca:
@pgom if you want to work on this, you can create a new PR, and I'll close this one.

iamandrewluca:
It seems this is harder then I thoughth.

jamuhl:
@iamandrewluca If ok for you I like to keep this open - even if you're not going to finish this PR. If getting to it I might take a step into this...

iamandrewluca:
Ok, no problems. 🙂

jamuhl:
@iamandrewluca can you help me with some test cases that should be covered?

Currently got this working:

```js
    <Trans
      defaults="hello <italic>beautiful</italic> <bold>{{what}}</bold>"
      values={{ what: 'world' }}
    />
```

jamuhl:
@iamandrewluca can you please check https://github.com/i18next/react-i18next/commit/f459740fb8e8cf80f00b541c35fbb5f31363fc8b (current master)

hopefully that solves the requested feature?

iamandrewluca:
Taking a look

iamandrewluca:
LGTM
@jamuhl I almost forgot the context, it looks so complicated. :slightly_smiling_face:

jamuhl:
agree...the Trans component once was rather simple...but with all those cases and the 2 modes (using children or props.components) grown to be complicated.

iamandrewluca:
I asked above what are some restrictions that `Trans` props should have between them, how can they be combined, and what is forbidden.

Having some strong rule in place how bellow props can be combined, we can make `Trans` implementation details more simple

`i18nKey`, `components`, `children`, `values`, `defaults`

jamuhl:
published eact-i18next@11.6.0 should now support object for props.components

jamuhl:
props.components = children
values = child (in form of objects)
defaults = nodesToString(children) (used in no t(i18nKey))
i18nKey -> explicit key (instead of using nodesToString(children) or defaults)

PR #860, opened 2019-05-29
Restore support passing the defaultNS as I18nextProvider prop 

alexbakum:
During react-i18next evolved the support (introduced at i18next/react-i18next#478) of `defaultNS` prop has been dropped from I18nextProvider. So attempting to restore it.
We're still looking forward to share same instance of i18next between different provider instances and hooks.
```javascript
const i18n = /* some i18next instantiating */
const Translated = () => {
  const {t} = useTranslation();

  return (<div>{t('the_key')}</div>);
};
const FirstWrapper = () => (
  <I18nextProvider i18n={i18n} defaultNS="firstOne">
    <Translated />
  </I18nextProvider>
);
const SecondWrapper = () => (
  <I18nextProvider i18n={i18n} defaultNS="secondOne">
    <Translated />
  </I18nextProvider>
);
```

### DIFF
--- a/src/I18nextProvider.js
+++ b/src/I18nextProvider.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import { I18nContext, usedI18nextProvider } from './context';
 
-export function I18nextProvider({ i18n, children }) {
+export function I18nextProvider({ i18n, defaultNS, children }) {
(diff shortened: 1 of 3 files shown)


jamuhl:
thank you...was published in react-i18next@10.11.0

PR #861, opened 2019-05-31
Fix wording

nlitwin:
- fix the grammar and wording of the heading in `README.md`
- this change also mirrors the language used [in the docs](https://github.com/i18next/react-i18next-gitbook/blob/master/README.md#what-does-my-code-look-like)


jamuhl:
thank you

PR #872, opened 2019-06-17
Update I18nextProvider.js

keustma:
Because of missing migration documentation (docs added in: https://github.com/i18next/react-i18next-gitbook/pull/53) in `next-i18next` which uses `react-i18next` a bug was introduced which prevented the normal working of the `Link` component. Cfr: https://github.com/isaachinman/next-i18next/issues/348

This PR adds a warning when a user tries to use a removed context prop.

### DIFF
--- a/src/I18nextProvider.js
+++ b/src/I18nextProvider.js
@@ -1,6 +1,12 @@
 import React from 'react';
 import { I18nContext, usedI18nextProvider } from './context';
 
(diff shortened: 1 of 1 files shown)


jamuhl:
Think this is not needed - it's now documented and beside next.js use case most moved already to v10 (where those props where in most cases not needed anyway).

Issue #817, opened 2019-04-04
React.lazy usage with useTranslation hook


DevanB:
**Describe the bug**
I have a few pages that use `React.lazy`. It seems that on those pages, my translations aren't loaded (or maybe are loaded after the component renders).

I've tried several variations of setting `react: { useSuspense: false}` and `useTranslation('translations'`, {useSuspense: false})` but have had no success getting translations to display. 

I'm not sure if this is a  bug, or a misconfiguration on my part.

**Occurs in react-i18next version**
10.6.1

**To Reproduce**
1. Use the following config:
`i18n.ts`
```
import i18n from 'i18next';
import LanguageDetector from 'i18next-browser-languagedetector';
import Backend from 'i18next-xhr-backend';
import {initReactI18next} from 'react-i18next';

i18n
  .use(Backend)
  .use(LanguageDetector)
  .use(initReactI18next)
  .init({
    defaultNS: 'translation',
    fallbackLng: 'en',
    interpolation: {
      escapeValue: false, // not needed for react as it escapes by default
    },
    load: 'languageOnly',
    saveMissing: true,
  });

export default i18n;
```

Optionally add `react: { useSuspense: false}`

2. Lazy load a component that uses the `useTranslation` hook:

```
const NotFound = React.lazy(() => import('./NotFound'));
```

```
import React from 'react';
import {useTranslation} from 'react-i18next';

const NotFound: React.SFC = () => {
  const [t] = useTranslation();
  return <h1 data-testid='not-found-heading'>{t('not-found')}</h1>;
};

export default NotFound;
```
Optionally use `useTranslation('translations'`, {useSuspense: false})`

**Expected behaviour**
I expect to see my translations, not the key of the translation.

**OS (please complete the following information):**

- Device: 2016 MBP 13" macOS 10.14.4
- Browser: Firefox Developer Edition 67.0b6

jamuhl:
Could you provide a code sandbox for reproduction? Not yet sure - never used React.lazy yet - so not sure what it does under the hood...but as it's react official i guess it should work with hooks out of the box...

DevanB:
Sure! Let me get one together for you. 😄

jamuhl:
thank you

DevanB:
Arg. I can't reproduce it in a CodeSandbox because the issue happens at a very specific point in time.

[...]

PR #878, opened 2019-06-20
Reload translations whenever namespaces passed to useTranslation() change

nicolasschabram:
When dynamically changing the `namespaces` passed to `useTranslation()` over time, they are now reloaded and trigger a re-render. This reimplements the behavior first introduced in #523 as it was likely broken in the release of the new hooks-based API.

See https://github.com/i18next/react-i18next/issues/817#issuecomment-503650139 and #520 for context.

### DIFF
--- a/src/useTranslation.js
+++ b/src/useTranslation.js
@@ -74,7 +74,7 @@ export function useTranslation(ns, props = {}) {
       if (bindI18nStore && i18n)
         bindI18nStore.split(' ').forEach(e => i18n.store.off(e, boundReset));
     };
-  }, []); // define props to trigger using effect on rerender (none here)
+  }, [namespaces.join()]); // re-run effect whenever list of namespaces changes
 
   const ret = [t.t, i18n, ready];
   ret.t = t.t;


src/useTranslation.js
```diff
@@ -74,7 +74,7 @@ export function useTranslation(ns, props = {}) {
       if (bindI18nStore && i18n)
         bindI18nStore.split(' ').forEach(e => i18n.store.off(e, boundReset));
     };
-  }, []); // define props to trigger using effect on rerender (none here)
+  }, [namespaces.join()]); // re-run effect whenever list of namespaces changes
```
nicolasschabram: I thought it might be safer not to rely on object identity in case people pass inlined arrays like `useTranslation(["x", "y"])` that are recreated on every render. That's why I'm turning the `namespaces` array into a string first.

jamuhl: yes...saw that...think that's rather clever

jamuhl:
was published in react-i18next@10.11.2

PR #883, opened 2019-06-24
Fix for error accessing Trans node children in Preact

ascorbic:
Preact throws a deprecation error if we try to access `vnode.children`. This PR switches the test around, so that we use `node.props.children` if it exists, rather than checking `node.children` first.


ascorbic:
Closing as I missed the test failures

PR #911, opened 2019-07-25
Add translation alias

danielmockaitis:
I've noticed a lot of development teams like to rename the `t` to` translate`.

### DIFF
--- a/src/withTranslation.js
+++ b/src/withTranslation.js
@@ -6,10 +6,10 @@ export function withTranslation(ns, options = {}) {
   return function Extend(WrappedComponent) {
     function I18nextWithTranslation(props, ref) {
       const [t, i18n, ready] = useTranslation(ns, props);
-
       const passDownProps = {
         ...props,
         t,
+        translate: t,
         i18n,
         tReady: ready,
       };


jamuhl:
And others like to name it `_` or `__` or `tr` or ... --> guess you get the point...i18next has `t`

If you prefer `translate` just create a own component `withTranslation` (in the end it's just a wrapper around `useTranslation` where you can `const [whatEverILikeTToBeCalled] = useTranslation();`

Silpaa:
Hi Jamuhl,
I am looking for the same thing. I want to rename it to translate. Which is more readable to my team.
Can you provide a sample code/example with withTranslation

jamuhl:
@aplis1 make a copy of https://github.com/i18next/react-i18next/blob/master/src/withTranslation.js and rename t to translate...

`translate: t,` https://github.com/i18next/react-i18next/blob/master/src/withTranslation.js#L12

aplis1:
Thanks @jamuhl . I will try and let you know

aplis1:
I am getting Error: withI18nextTranslation(App) suspended while rendering, but no fallback UI was specified.****
Can you please provide a codesandbox working solution.

adrai:
@aplis1 can you provide a codesandbox showing the error?
You probably disabled suspense - so you're responsible to not render / access t before ready: https://react.i18next.com/latest/usetranslation-hook#not-using-suspense

aplis1:
@adrai - I am new to react and i18next. Please have a look at this-
https://codesandbox.io/s/new-snow-fkjz6?fontsize=14&hidenavigation=1&theme=dark

adrai:
@aplis1  maybe there is a better example, but this should work: https://codesandbox.io/s/blissful-mountain-d8581?fontsize=14&hidenavigation=1&theme=dark

aplis1:
Thank you @adrai

aplis1:
@adrai why do you have to add wrapper class again.

adrai:
because you are using both in your code: useTranslation and withTranslation

PR #912, opened 2019-08-01
fix types for use()

rosskevin:
Depends on https://github.com/i18next/i18next/pull/1296

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
     "eslint-plugin-jsx-a11y": "6.2.1",
     "eslint-plugin-react": "7.12.4",
     "husky": "^1.3.1",
-    "i18next": "^15.0.7",
+    "i18next": "^17.0.8",
     "jest": "24.1.0",
     "jest-cli": "24.1.0",
     "lint-staged": "^8.1.3",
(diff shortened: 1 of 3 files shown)


rosskevin:

jamuhl:
published in  react-i18next@10.11.5

kuzvac:
Hi @rosskevin, today i updated all our i18next dependencies and found that types for `t` is lost, if i use withTranslation HOC. After short investigation i found that this type of import `import * as i18next from 'i18next';` is broke types definition borrowed from i18next. 
image
201909198ojz8

But if change this import just to `import i18next from 'i18next';` everything is working properly. 
I can make PR for replace `import * as i18next from 'i18next';` to `import i18next from 'i18next';` :)

rosskevin:
@kuzvac please see https://github.com/i18next/react-i18next/issues/931#issuecomment-526636217 or file a new PR with a breaking test case. This is thoroughly tested.

kuzvac:
@rosskevin thanks. hmmm... react-i18next test is working properly with our tsconfig & tslint files. Looks like it's problem with our code/dependencies :)

kuzvac:
@rosskevin I found why the error occurs in my case :) 
I do clean install, compare version of i18next modules, and found that, in version that doesn't work, i have `i18next^15.1.3` and `react-i18next^10.12.5`
What a mess, shame on me 😅

PR #917, opened 2019-08-09
ICU: Trans macro will parse defaults as alternate to children

redbugz:
Fix for https://github.com/i18next/i18next/issues/1295
Adds support for ICU message syntax inside the `defaults` attribute
This allows the Trans macro to be used with linters and other code analysis
tools that expect valid JSX syntax as children, by using the defaults
prop instead of children for the message.

Also adds support fo the `<SelectOrdinal>` tag
Fixes a bug where `Trans` was imported from 'react/i18next' instead of 'react-i18next'


redbugz:
I'm not sure if I am supposed to commit react-i18n.js and react-i18next.min.js or not.

redbugz:
Would you like me to split up the icu.macro.js file to get the line count under 250 to fix the codeclimate issue?

jamuhl:
@redbugz no you can ignore the code climate for this one - I'm ok with that

I will merge this later today and publish an updated version...thank you for contributing

jamuhl:
published in react-i18next@10.12.0

jamuhl:

PR #918, opened 2019-08-12
Set ready flag to false when i18n instance has not been initialised

nickythorne:


jamuhl:
Does this make a difference? I mean this is an issue in code and I hope your app never ships with this problem of not having passed an i18n instance down there? So you get the warning anyway...ready: true|false makes no difference in this case...the userland code is broken when getting there.

nickythorne:
Setting {{read}} to false for this particular code flow (when i18n is not initialised) just makes sense. But I believe there is also a genuine use case for such behaviour... If the initialisation of the i18n instance is deferred (the reason for which is irrelevant), relying on the correctness of this flag could be quite useful to enable client applications to display some alternative component before it is initialised, for example, displaying placeholders.

An example: 

`const Header: React.FC = () => {

    const { t, ready } = useTranslation();

    return (
        <Container>
            <Placeholder ready={ready}>
                <Title>{t("header.title")}</Title>
                <SubTitle>{t("header.subtitle")}</SubTitle>
            </Placeholder>
        </Container>
    )
};`

jamuhl:
published in react-i18next@10.12.1

Issue #931, opened 2019-08-29
Typings broken


zombiQWERTY:
**Describe the bug**
When using with typescript there is a bug with index.d.ts

**Occurs in react-i18next version**

npm module version: 10.12.2

**To Reproduce**

1. Make ts app with react-i18next
2. Write types for props such as:
`type RouteMapProps = ReturnType<typeof mapStateToProps> & typeof mapDispatchToProps & OwnProps & WithTranslation`
3. See error

**Expected behaviour**
No type errors

**Additional context**
A problem with first line index.d.ts file.
A month ago there was this line:
`import i18next from 'i18next';`
But now there is:
`import * as i18next from 'i18next';`

Version 10.11.4 works fine

rosskevin:
Please post the contents of your `tsconfig.json`

zombiQWERTY:
https://gist.github.com/zombiQWERTY/07f508b65aeea4ec91a8e7521883e43f
Do you think the problem with "moduleResolution" directive?

rosskevin:
No, my guess is you specified typeRoots which I believe precludes node_modules. You want to add a path to include instead and omit typeRoots

rosskevin:
Ok, I have had time to look at this further.  We did not have the paranoid ts tests like i18next, so I added them in #932.

> A month ago there was this line:
> import i18next from 'i18next';
> But now there is:
> import * as i18next from 'i18next';

Correct.  `import * as` will _always_ work regardless of your `esModuleInterop` setting, therefore it change for compatibility.

Given that all tests work (no changes needed for the extra tests in #932), I'm pretty confident that @zombiQWERTY has a local misconfiguration that I referenced above.

If you do have a problem, please PR a breaking test and tag me.

PR #932, opened 2019-08-30
Add ts esModuleInterop false and true tests

rosskevin:
Prove that `esModuleInterop` both `false` and `true` is working for types.

Related: #931


rosskevin:
@jamuhl no release necessary - just more tests.

Issue #934, opened 2019-09-07
Change in the order of Hooks


PsychedMagnet:
**Describe the bug**
The rule of Hooks is broken in the useTranslation hook.

**Occurs in react-i18next version**
I believe the commit that first introduced it is `0fafb4563a9c852792bd6851b45008d1315566f0`.
[Here is the exact place in the current version](https://github.com/i18next/react-i18next/blob/f41a24bddfaa24f96e706cd6af5744da337d4696/src/useTranslation.js#L14-L16)

**To Reproduce**
N/A

**Expected behaviour**
All hooks should be called every render in the same order.

**Screenshots**
N/A

**OS (please complete the following information):**
N/A

**Additional context**
I am not familiar with the internal workings of i18next, so I can not give exact steps to reproduce, but I get the warning when I switch pages in Gatsby, although it probably happens in other situations too.

PsychedMagnet:
I made a Pull Request, which I think fixes it the problem - #935

PsychedMagnet:
I am closing this issue as my PR has been merged and the issue has been resolved.

PR #935, opened 2019-09-07
Fix hooks being called conditionally in useTranslation

PsychedMagnet:
In the `useTranslation` hook, `useContext` is called conditionally, which breaks the Rule of Hooks, so it may cause problems in the future. This fixes it by calling `useContext` on every render, just not always using the result from it.

### DIFF
--- a/src/useTranslation.js
+++ b/src/useTranslation.js
@@ -11,8 +11,9 @@ import { warnOnce, loadNamespaces, hasLoadedNamespace } from './utils';
 export function useTranslation(ns, props = {}) {
   // assert we have the needed i18nInstance
   const { i18n: i18nFromProps } = props;
+  const ReactI18nContext = useContext(I18nContext);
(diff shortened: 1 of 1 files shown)


PsychedMagnet:
This is a fix for #934

jamuhl:
Guess that will work - and also not against a merge.

But I would love to know how this can happen. The result from `getHasUsedI18nextProvider()`  is always the same depending of using `I18nextProvider` or not, so how could there be a change of order? Only way i could think of is nesting `I18nextProvider` inside a `useTranslation|withTranslation` ?!?

PsychedMagnet:
Oh, you are right. There is a goof up on my part - I was calling `useTranslation` above `I18nextProvider`. I will definitely fix this, thank you! :)

Anyways, I think unconditionally calling `useContext` is the better option as a good practice either way.

jamuhl:
published in react-i18next@10.12.3

Issue #937, opened 2019-09-09
icu.macro for Plural/SelectOrdinal does not handle count prop of name count, or expressions


redbugz:
**Describe the bug**
I am working on a PR to fix this, but wanted to start a discussion about the best way to handle it.

When using the non-`Trans` icu macros `Plural/SelectOrdinal` (and `Select` with the `switch` prop), the `count` prop is put into `values` during transformation, and since `count` is handled specially by the normal `Trans` component, it does not render correctly.
This means you cannot use a variable named `count` in `Plural` or `SelectOrdinal`

Also, it must be a standalone variable, no expressions are allowed. If you use an expression in the `count` or `switch` props, you get `TypeError: react-i18next/icu.macro: Property name expected type of string but got null`

**Occurs in react-i18next version**
all versions I believe. I don't think this has ever worked, it has either never come up, or always got worked around by introducing local variables

**To Reproduce**
Use `Plural` with a variable named `count`

```
<Plural
          i18nKey="somekey"
          count={count}
          $0="There are no items."
          one="There is <Link to='items'># item</Link>."
          other="There are <Link to='items'># items</Link>."
        />
```
This results in the `#` being replaced with `NaN` because `count={count}` above is transformed into `<Trans values={{count}}...` and `Trans` requires that `count` be a special prop. So you end up with opts like this:
```
{ count: undefined, values: {count: count} }
```
Since `count` is undefined, you get `NaN` instead of using the `count` in values.

OR use `Plural` with `count` as an expression
```
<Plural
          i18nKey="items.link.string" // optional key
          count={index+1}
          $0="There are no items."
          one="There is <Link to='items'># item</Link>."
          other="There are <Link to='items'># items</Link>."
        />
```
This results in the TypeError from above.

**Expected behaviour**
The `count` prop of `Plural/SelectOrdinal` and the `switch` prop of `Select` can be an expression or a variable named `count` 

**Additional context**
For the variable named `count`, the solutions I see are either to transform `count` into the prop named `count`
`<Plural count={count} --> <Trans count={count}`
or to allow the `count` prop to come from values if it isn't defined as it's own prop.

For not allowing expressions, we can either just document that limitation, or transform it correctly:
`<Plural count={index+1} --> <Trans count={index+1}`
OR if we decide to allow count to come from values as a fallback:
`<Plural count={index+1} --> <Trans values={{count:index+1}}`

[...]

PR #939, opened 2019-09-10
ICU: Fixes macro to support count prop and expressions better

redbugz:
Fixes for #937
Allows count prop to use a variable also named count
Add support for expressions in the count and switch props
Handles the values prop better
Allows macros to be self-closing without children
Trims newlines and whitespace around code formatted with prettier or other formatting tools

If the `count` prop is also named `count`, it copies it to the `count` prop of `Trans`:
```
<Plural count={count} ---> <Trans count={count}
```
If `count` is an expression it also copies it to `Trans` as the `count` prop:
```
<Plural count={index+1} ---> <Trans count={index+1}
```

Also, if code is formatted with prettier, the extra whitespace was getting into the translations, so now trims the whitespace back to normal

If you had a values prop and a count, sometimes they didn't merge correctly. I believe that is fixed now.
```
<Plural
  count={number}
  values={{ prop1: value1, prop2: value2 }}
/>
// this transforms to:
<Trans values={{ number, prop1: value1, prop2: value2 }} />
```

The above also works for `SelectOrdinal` and `Select` with the `switch` prop.


redbugz:
Somehow in my merge/rebase of master, I ended up with logs of extra commits, but the changed files is correct. If you would like me to clean up the commits, I can do the PR again.

icu.macro.js
```diff
@@ -137,26 +141,33 @@ function selectAsJSX(parentPath, { attributes }, babel) {
         // copy the i18nKey
         mem.attributesToCopy.push(attr.node);
       } else if (attr.node.name.name === 'switch') {
-        // take the switch for plural element
-        mem.values.push(toObjectProperty(attr.node.value.expression.name));
-        mem.defaults = `{${attr.node.value.expression.name}, select, ${mem.defaults}`;
+        // take the switch for select element
+        let exprName = attr.node.value.expression.name;
+        if (!exprName) {
+          exprName = 'selectKey';
```
redbugz: I couldn't use `switch` because it is a reserved word. I'm open to alternate names for this.

It is used when `switch` is an expression.

```
<Select
  switch={'fe'+'male'} // very contrived example
  ...
/>
// This transforms to:
<Trans values={{ selectKey: 'fe'+'male' }} defaults="{selectKey, select, ....} />
```
I don't think this will be used very often, but it was the same logic as `Plural` and `count` so I included it for completeness.

icu.macro.js
```diff
@@ -85,39 +68,54 @@ function pluralAsJSX(parentPath, { attributes }, babel) {
...
-        const children = attr.node.value.expression.children;
+        const children = attr.node.value.expression.children || [];
         const thisTrans = processTrans(children, babel, componentStartIndex);
 
         let pluralForm = attr.node.name.name;
         if (pluralForm.indexOf('$') === 0) pluralForm = pluralForm.replace('$', '=');
 
         mem.defaults = `${mem.defaults} ${pluralForm} {${thisTrans.defaults}}`;
         mem.components = mem.components.concat(thisTrans.components);
-        mem.values = mem.values.concat(thisTrans.values);
```
redbugz: this was causing problems with duplicate keys in `values`, and I'm not sure I understand the use case and all the test cases I needed are passing, so I just removed it.
If we need to support this, then I can add code to dedupe these from `values`.

icu.macro.js
```diff
@@ -137,26 +141,33 @@ function selectAsJSX(parentPath, { attributes }, babel) {
...
         mem.defaults = `${mem.defaults} ${attr.node.name.name} {${attr.node.value.value}}`;
       } else if (attr.node.value.type === 'JSXExpressionContainer') {
         // convert any Trans component to select option extracting any values and components
-        const children = attr.node.value.expression.children;
+        const children = attr.node.value.expression.children || [];
         const thisTrans = processTrans(children, babel, componentStartIndex);
 
         mem.defaults = `${mem.defaults} ${attr.node.name.name} {${thisTrans.defaults}}`;
         mem.components = mem.components.concat(thisTrans.components);
-        mem.values = mem.values.concat(thisTrans.values);
```
redbugz: see comment above for similar code

icu.macro.js
jamuhl: looks ok for me

jamuhl:
Will check this later today and merge/publish

jamuhl:
published in react-i18next@10.12.4

Issue #711, opened 2019-02-02
Another approach for SSR with multiple namespaces


klis87:
**Is your feature request related to a problem? Please describe.**
Current suggestion for SSR is in my opinion not complete and optimal, especially when having multiple namespaces. In https://react.i18next.com/misc/serverside-rendering#loadnamespaces-helper it is mentioned we could use `match` from React-router to know which namespaces are needed. This is really problematic because:
- it works only with legacy router version 3
- not everyone uses react-router, especially when using redux and pure redux routers like `redux-first-router`
- even if you use React-router 3, then still this method assumes that children of Route components won't need dedicated namespaces, which forces us to decorade only route components with namespaces hocs, which is unflexible and often not optimal

**Describe the solution you'd like**
I suggest a simpler solution for SSR:
- we should mention in the docs about possibility to preload all namespaces for all languages at server boot, on server it has to be done only once anyway, so it doesn't hurt to do it immediately
- then, `withNamespaces` for a given React component on the server won't be actually used to load a given namespace for a given language  (yes, it will load namespaces but this would be cached already), it would be used to know which namespaces are needed to be passed to the client in `initialI18nStore`

How we could achieve it? For each request we could create an object like `set` or `[]` (set would be better but maybe you would prefer array as set is not implemented in every browser) and pass it to provider like:
```js
const usedNamespaces = new Set();
<I18nextProvider i18n={req.i18n} initialLanguage={language} usedNamespaces={usedNamespaces}>
```

Now, `usedNamespaces` could be added to i18next context, and in proper i18next hocs/components, like `withNamespaces`, `usedNamespaces` set could be mutated with used namespace/namespaces in a given component, for example `usedNamespaces.add('someNamespaceName')`.

Then, after components are rendered, we could read `usedNamespaces` to see which namespaces were used for this specific request. Then, we can used this info to pass optimal  `initialI18nStore` to the client.

**Describe alternatives you've considered**
- using one namespace - not scalable
- passing all namespaces to `initialI18nStore` - defeats the purpose of using namespaces in the first place
- keeping separate configuration which namespaces are required for all routes - requires keeping separate config, while this can be easily computed as suggested automatically

**Additional context**
N/A

jamuhl:
That's not a suggestion - it's one of the things you can do. It's impossible to write a documentation for every possible combination of things existing out there. So all the SSR stuff is giving hints, directions how to approach the problem.

[...]

PR #945, opened 2019-09-13
Update types for reportNamespaces

skovy:
I'm working with SSR and stumbled across this issue and need access to only the used namespaces: https://github.com/i18next/react-i18next/issues/711.

As it appears, the suggestion is to use `i18next.reportNamespaces.getUsedNamespaces()`?

However, the types do not currently reflect that. 

### Previously

```typescript
import * as i18next from 'i18next';

i18next.reportNamespaces.addUsedNamespaces(['translation']);
//      ^^^^^^^^^^^^^^^^
// Property 'reportNamespaces' does not exist on type 'i18n'
i18next.reportNamespaces.getUsedNamespaces();
//      ^^^^^^^^^^^^^^^^
// Property 'reportNamespaces' does not exist on type 'i18n'
```

### After this Change

```typescript
import * as i18next from 'i18next';

i18next.reportNamespaces.addUsedNamespaces(['translation']); // valid
i18next.reportNamespaces.getUsedNamespaces(); // valid, returns: `string[]`
```

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -7,8 +7,6 @@ export type Namespace = string | string[];
 
 export function setDefaults(options: i18next.ReactOptions): void;
 export function getDefaults(): i18next.ReactOptions;
-export function addUsedNamespaces(namespaces: Namespace[]): void;
-export function getUsedNamespaces(): string[];
 export function setI18n(instance: i18next.i18n): void;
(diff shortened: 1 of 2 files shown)


rosskevin:
Thanks @skovy.

jamuhl:
published in react-i18next@10.12.5

PR #947, opened 2019-09-24
Use count from `values` object passed to Trans

k4r1:
If a `count` is specified in the `values` object passed to Trans, we should use it, even if the `count` property is not provided.
https://github.com/i18next/i18next/issues/1330

By changing the order of the construction of `combinedTOpts`, this should ensure that the behaviour is as expected.


jamuhl:
published in react-i18next@10.13.0

Issue #954, opened 2019-09-30
Backward compatibility issue


mlecoq:
**Describe the bug**
I am using react i18next with typescript

I am using useTranslation to get current language

```
    const { t, i18n } = useTranslation();
...
  
  const language = i18n.language
```

I have upgraded my project and then I have a compilation issue on i18n.language

**Occurs in react-i18next version**
Occurs with #945 - works properly with version 10.12.2 

Is there any workaround ?

Thanks !

rosskevin:
@mlecoq I have proven this to be working in #956.  

Since you did not provide the ts error you are receiving, and no other details of your setup, I have to assume that this is a local problem.

mlecoq:
the ts error is : 

`Property 'language' does not exist on type 'i18n'.ts(2339)` 

I only have access to property `reportNamespaces`

mlecoq:
When I look to type definition, 

```
declare module 'i18next' {
  interface i18n {
    reportNamespaces: ReportNamespaces;
  }
}
```

rosskevin:
^^^ `declare module` code is a typescript augmentation of the existing definition from `i18next`.

Perhaps your dependencies are out of date, or you are configuring tsconfig in a problematic way.

https://github.com/i18next/react-i18next/pull/956/files proves this works, otherwise the tests would not pass.

mlecoq:
Yes, you're right. I have found the issue: i18next was still in version 15.0.9. After upgrading to 17.0.4, the problem is fixed. 
Sorry !

mlecoq:
Yes you're right, i18next was in version 15.1.3 

I have upgraded it to 17.0.4 and It works properly

Sorry !

mlecoq:
Maybe an upgrade of peer dependencies ? 
  "i18next": ">= 17.0.0"

rosskevin:
Sure, please PR that @mlecoq and tag me, thank you.

PR #956, opened 2019-09-30
Test exposed i18n types

rosskevin:
Try to assert issue reported in #954


rosskevin:
No release necessary.

PR #957, opened 2019-10-02
update peer dependency to fix typescript issue

mlecoq:
PR following discussions on #954 with @rosskevin


rosskevin:
Thank you @mlecoq 

jamuhl:
Guess this can be released with the next version (no extra publish needed) - needed peerDependency is also only adjusted for typescript - js user still can go back to v14 with no issue at all

Issue #958, opened 2019-10-03
HOC `withTranslation` doesn't copy static attributes and methods over


percypyan:
**Describe the bug**
The statics of WrappedComponent aren't passed over the `withTranslation` HOC.

**Occurs in react-i18next version**
Noticed in version 10.13.1, but I suppose it's this way since the introduction of `withTranslation` HOC to replace the `translate` one.

**To Reproduce**
Simply define some statics on a component, then try to use one those statics on the WrappingComponent.

I've already done a Pull Request (#959) with a non-regression test, you can refer to it. 

**Expected behaviour**
We expect the statics to be passed over.

jamuhl:
https://react.i18next.com/latest/withtranslation-hoc#hoist-non-react-statics it's expected behaviour

percypyan:
Yep, my bad. I've should have checked.

PR #959, opened 2019-10-03
Update withTranslation HOC to pass statics over

percypyan:
This Pull Request is fixing the issue #958.

React Documentation: https://reactjs.org/docs/higher-order-components.html#static-methods-must-be-copied-over

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.3.1",
+    "hoist-non-react-statics": "^3.3.0",
     "html-parse-stringify2": "2.0.1"
   },
   "devDependencies": {
--- a/src/withTranslation.js
+++ b/src/withTranslation.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import hoistNonReactStatics from 'hoist-non-react-statics';
 import { useTranslation } from './useTranslation';
 import { getDisplayName } from './utils';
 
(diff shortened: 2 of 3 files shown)


jamuhl:
It's already discussed here: https://react.i18next.com/latest/withtranslation-hoc#hoist-non-react-statics

I won't add that module as it adds to much weight to react-i18next module.

percypyan:
Oh. I'm sorry, I didn't look here, my bad.
I jump from v6 to v10 and I had in mind that `withTranslation` was the new `translate`, that use to hoist the statics.

I understand here the need to keep the package light but don't you think that a package like yours, whose purpose is to provide tools to use i18next with React the right way should provide a fully functional HOC ? Even if it mean add 37ms of loading time at 30 kB/s ?

I'll close the PR, and will write a wrapper to the HOC in my apps. Sorry for the false issue.

jamuhl:
Depends on to whom you listen more...the once that want convenience...or the once that want light weight...currently is i get this one not reported a lot anymore I guess people are ok with hoisting in userland if needed.

percypyan:
Of course, you can see it this way, and when you are, let the hoist to the user make sense. But it's not for convenience I think you should do it directly in the HOC, it's for consistency: most of the HOC pass the statics over. The React documentation give multiple ways to deal with the statics and of course there is one where it's not the responsibility of the HOC, but this one come in last.

Still thank you for the module, even if it's not perfect to me ;)

PR #961, opened 2019-10-03
Fix import of i18next so that types are extended

CaptEmulation:
When trying to use the `18next` instance provided by `useTranslation` hook in a typescript app there missing definitions. The problem was the import of `i18next` in the typescript definition. Because `esModuleInterop` is enabled then we should use the default export. If you look at i18next's own typescript definition then we can see it is a CommonJS "interop default" export with no named exports:
https://github.com/i18next/i18next/blob/master/index.d.ts#L1008

Allows the following code to pass type check:

```
  const [t, i18n] = useTranslation('common')

  useEffect(() => {
   // otherwise fails with:
   //     Property 'on' does not exist on type 'i18n'.  TS2339
   i18n.on('languageChanged', doThing)
  }, [])
```

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -1,4 +1,4 @@
-import * as i18next from 'i18next';
+import i18next from 'i18next';
 import * as React from 'react';
 
 type Omit<T, K> = Pick<T, Exclude<keyof T, K>>;


rosskevin:
>  Because `esModuleInterop` is enabled 

this is not always true, this library is used by those with and without `esModuleInterop`, and by default this is false.

`import * as` works for both values of `esModuleInterop`, therefore it is the default.  These are tested in `test/typescript` for `esModuleInterop: true` and in `test/typescript/nonEsModuleInterop` for `esModuleInterop: false` (the default value).  If you are finding a problem with types, please PR a breaking test in the relevant folder related to your use of `esModuleInterop`.

Does that make sense @CaptEmulation?  Unfortunately this is a nuanced issue that is harder than it needs to be!

CaptEmulation:
@rosskevin Thank you for taking the time to explain all of that! With the test structure as-is I am not able to duplicate my problem. My hunch is that I would need create a test that has root package.json as another package.json dependency and run an install. AFAICT when I have a `esModuleInterop` project consuming `react-i18next` that no matter whether I have `i18next` installed myself when I try to resolve the `i18next` instance provided by `useTranslation` that I only get the https://github.com/i18next/react-i18next/blob/master/src/index.d.ts#L26-L30 as my i18next declaration because I assume from that file my `esModuleInterop` project is not able to resolve the library any further.

I've already forked and published to unblock me, but I would prefer not to have to maintain said fork so if I can I will see that in light of the information you provided there is a way to satisfy (and test for!) both variation of consuming `react-i18next` _as a module_ with and without `esModuleInterop`

PR #965, opened 2019-10-12
Add cross-env

tisoap:
### Features

Add [cross-env](https://www.npmjs.com/package/cross-env), which allows for scripts to be executed on Windows systems.


tisoap:
About the currently failing test, I double checked by running `npm start` on the `master` branch in a linux machine, and the error persisted. React displays this message:

```
Warning: Function components cannot be given refs. Attempts to access this ref will
fail. Did you mean to use React.forwardRef()?
      
Check the render method of `ForwardRef(withI18nextTranslation(TestComponent))`.
    in TestComponent (created by ForwardRef(withI18nextTranslation(TestComponent)))
    in ForwardRef(withI18nextTranslation(TestComponent)) (created by WrapperComponent)
    in WrapperComponent
```

jamuhl:
why we need that? we build on mac and publish to npm - no need to build this on windows

jamuhl:
previous builds still run through with success - so the problem must be with adding crossenv

tisoap:
> why we need that? we build on mac and publish to npm - no need to build this on windows

So that future contributors that happen to be working on a Windows machine (like me right now) can still run the test scripts without problems.

> previous builds still run through with success - so the problem must be with adding crossenv

I did a new clone of this repo in my notebook running linux, and running `npm install` and `npm test` on the `master` branch yields the same error. Try to confirm if this happens to you too in your local machine. Maybe since this project is ignoring `package-lock.json` for some reason, a `npm install` on a new clone is installing different dependencies the CI was using on previous builds?

tisoap:
The tests are passing now. That was fast! :slightly_smiling_face:

jamuhl:
thank you for the contribution

PR #966, opened 2019-10-12
Fix linter errors

tisoap:
While reading the source code, I noticed the project still had some linter warnings showing up, so I decided to address them before anything. The list of changes:

- Added `/test` folder to `test:lint` script
- Fixed all errors and warnings produced by `npm run test:lint`
- Disabled `react/jsx-props-no-spreading` rule, since it is being used by `withTranslation.js`
- Updated `eslint` dependencies, which in turn removed a linter warning from `useTranslation.loading.spec.js`


jamuhl:
thank you

PR #969, opened 2019-10-18
Add t function to TransProps types

valerii-cognite:
It seems `t` function is currently missing from types, let's add it back


valerii-cognite:
@rosskevin added

rosskevin:

jamuhl:
published in react-i18next@10.13.2

Issue #974, opened 2019-10-22
Trans component and nesting


reintroducing:
Given the following JSON:
`"terms": "By continuing, I accept the <0>Terms of Service<2>, opens PDF file in a new window</2></0> and <1>Privacy Policy<2>, opens PDF file in a new window</2></1>.",`

I was expecting to be able to do this:
```js
<Trans
    i18nKey="user:login.terms"
    components={[
        /* eslint-disable react/jsx-key */
        <a
            className="underlined-link"
            href={tosUrl}
            target="_blank"
            rel="noopener noreferrer"
        />,
        <a
            className="underlined-link"
            href={privacyUrl}
            target="_blank"
            rel="noopener noreferrer"
        />,
        <span className="sr-only" />,
        /* eslint-enable react/jsx-key */
    ]}
/>
```

and output this:
`By continuing, I accept the Terms of Service and Privacy Policy.`

However, the output was this:
`By continuing, I accept the  and .`

The reason I want to be able to do this is because the `<2>, opens PDF file in a new window</2>` portion is supposed to come out as `<span className="sr-only">, opens PDF file in a new window</span>` in the final HTML where `sr-only` is a class tied to hiding content visually but allowing screen readers to still read it.

What is the proper way to handle something like this, without flat out putting that span tag markup in the translation JSON files?

jamuhl:
I won't go through explaining it all again - the answer can be found here: https://github.com/i18next/react-i18next/issues/953#issuecomment-538066505

And like said here (reading the full issue first) https://github.com/i18next/react-i18next/issues/953#issuecomment-539992389 - I completely agree with @tisoap and am open to accept a PR enabling that or even better would love to see this finally finished: https://github.com/i18next/react-i18next/issues/833 as having named tags and an object as components would be even more simple to use.

reintroducing:
@jamuhl I apologize for the seemingly repetitive question but that issue (which I missed, found some others that were similar but I could not get working based on solutions provided there) was the missing link for me. @tisoap solution in that thread was what opened it up for me to also understand how this should be used. Appreciate it.

Issue #953, opened 2019-09-30
Nested HTML element within Trans seems like not working


jackie-onai:
**Describe the bug**
Nested HTML element within Trans seems like not working

**Occurs in react-i18next version**
10.11.3

**To Reproduce**

i have this
```
<Trans i18nKey={"Are you sure you want to <0>delete</0> account  <0><1>{{name}}</1></0>  ?"} values={{name}} components={[<b key={0} />, <i key={1} />]} >
</Trans>               
```
{{name}} should be bold then italic, however, it becomes simply not shown.

jamuhl:
Does not work this way...inner `<0><1>` -> <1> must be inside b it's nesting not replacing by number

<Trans i18nKey={"Are you sure you want to <0>delete</0> account  <1>{{name}}</1>  ?"} values={{name}} components={[<b key={0} />, <b><i key={1} /></b>]} >
</Trans>

tisoap:
@jamuhl  I'm facing a similar problem, and I still don't understand how to properly use the `components` array, if they don't work by replacing the equivalent numbered tag. The documentation is not clear on this.

I need to map a large array of translations, with varying nesting of elements, and tried do something similar to this:

```jsx
const keys = [
  'test1', // "<0>Text inside paragraph</0>"
  'test2', // "<0>Text inside paragraph with <1>bold text</1></0>",
  'test3', // "just some <1>bold text</1>",
  'test4' // "<0>Text inside paragraph with <3>a link</3> for the user</0>"
]

// ...

<ul>
  {keys.map(key => (
    <li key={key}>
      <Trans
        i18nKey={key}
        components={[<p />, <strong />, <a href='#'></a>]}
      />
    </li>
  ))}
</ul>
```

Expected output:

```html
<ul>
  <li>
    <p>Text inside paragraph</p>
  </li>
  <li>
    <p>Text inside paragraph with <strong>bold text</strong></p>
  </li>
  <li>just some <strong>bold text</strong></li>
  <li>
    <p>Text inside paragraph with <a href='#'>a link</a> for the user</p>
  </li>
</ul>
```
Actual output:
```html
<ul>
  <li>
    <p>Text inside paragraph</p>
  </li>
  <li>
    <p></p>
  </li>
  <li>just some <strong>bold text</strong></li>
  <li>
    <p></p>
  </li>
</ul>
```

How could I achieve the "expected" output using the `<Trans>` component, or by any other means in `react-i18next`?

[...]

Issue #470, opened 2018-07-07
Error when nesting components without children using the defaults prop


beheh:
We've been using the new `defaults` strategy from #439 and been really happy with it. Unfortunately we're seeing errors when nesting components.

The following `Trans` component:

```jsx
<Trans
  defaults="<0>hello <1></1> {{what}}</0>"
  values={{ what: 'world'}}
  components={[<span>placeholder</span>, <br>placeholder</br>]}
></Trans>
```
throws the following error

```
TypeError: Cannot read property 'content' of undefined
  at src/Trans.js:78:86
    at Array.reduce (<anonymous>)
  at mapAST (src/Trans.js:62:21)
  at src/Trans.js:70:23
    at Array.reduce (<anonymous>)
  at mapAST (src/Trans.js:62:21)
  at src/Trans.js:70:23
    at Array.reduce (<anonymous>)
  at mapAST (src/Trans.js:62:21)
  at renderNodes (src/Trans.js:93:16)
```

I'm aware that right now I always need to add children to childrenless nodes, and that's fine in the toplevel case without nesting. I would like to be able to nest them though.

jamuhl:
not tested nesting but should be same as with having regular children - just the values get interpolated regular i18next way - so i guess it should be:

```js
<Trans
  defaults="<0>hello <1></1> {{what}}</0>"
  values={{ what: 'world'}}
  components={[<span>placeholder<br /></span> ]}
></Trans>
```

-> `br` must be nested in component at index 1

beheh:
Hm, while this works, the following doesn't:

```jsx
<Trans
  defaults="<0>prefix</0> <1>hello <2></2> {{what}}</1>"
  values={{ what: 'world'}}
  components={[<span>placeholder</span>, <span>placeholder<br /></span> ]}
></Trans>
```

We encounter this patterns in a few places of our codebase where `<0>` is the heading and `<1>` is a continued subheading.

I'm at a bit at a loss with regards to how the indexing works, maybe another example will help me to understand this better.

jamuhl:
<Trans
  defaults="<0>prefix</0> <1>hello <1></1> {{what}}</1>"
  values={{ what: 'world'}}
  components={[<span>placeholder</span>, <span>placeholder<br /></span> ]}
></Trans>

should work...think of it like <0>**<0>**prefix**</0>**</0> <1>**<0>**hello**</0>** <1></1><2>{{what}}</2></1>

jamuhl:
[...]

PR #981, opened 2019-10-27
wip: Indexed Components Prop

tisoap:
This PR alters the current behavior of `<Trans>` `components` prop, as to simplify the API in the case of nested components.

Currently, the `components` array must mimic the nesting used by the translation, and the numbered tags inside the translation "reset" at every new nesting level:

```jsx
<Trans
  defaults="<0>hello <0>there</0></0>" {/* Notice the two <0> tags */}
  components={[
    <span>
      placeholder<em>text</em> {/* <em> is inside <span> */}
    </span>
  ]}
/>

// output: <span>hello <em>there</em></span>
```

The proposed API makes it so the numbered tags inside the translation match the array index passed of `components`:

```jsx
<Trans
  defaults="<0>hello <1>there</1></0>"
  components={[
    <span key='0' />,
    <em key='1' />,
    {/* The `key` props are just there to avoid a warning from React for not
    using keys on an array of elements. But they can also be used to make
    easier to associate a component to where it will be used on the translation. */}
  ]}
/>

// output: <span>hello <em>there</em></span>
```

This should avoid confusion when using nested `components`  array, as it already happened before:

https://github.com/i18next/react-i18next/issues/470
https://github.com/i18next/react-i18next/issues/953
https://github.com/i18next/react-i18next/issues/974

**This PR introduces a breaking change.** Code written in the previous syntax will not work properly anymore. Using `list.map` inside the `components` prop array is no longer supported.

### DIFF
--- a/src/Trans.js
+++ b/src/Trans.js
@@ -88,13 +88,7 @@ export function nodesToString(startingString, children, index, i18nOptions) {
   return stringNode;
 }
 
-function renderNodes({
-  children,
-  targetString,
-  i18n,
-  i18nOptions,
-  combinedTOpts
-}) {
+function renderNodes({ children, components, targetString, i18n, i18nOptions, combinedTOpts }) {
   if (targetString === '') return [];
 
   // check if contains tags we need to replace from html string to react nodes
@@ -143,6 +137,11 @@ function renderNodes({
 
         if (typeof child === 'string') {
           mem.push(child);
+        } else if (components) {
+          const mappedChildren = mapAST(components, node.children);
+          const inner = mappedChildren.length === 0 ? null : mappedChildren;
+          if (child.dummy) child.children = inner;
+          mem.push(React.cloneElement(child, { ...child.props, key: i }, inner));
         } else if (hasChildren(child)) {
           const childs = getChildren(child);
           const mappedChildren = mapAST(childs, node.children);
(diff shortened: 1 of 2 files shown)


tisoap:
I did some mild refactoring before implementing the feature, which I guess _codeclimate_ picked up the changed lines and pointed code quality issues. Those quality issues where already inherent from the _Trans.js_ file, but breaking it down into separate files exposed them. I can undo those refactors if deemed necessary for the PR approval, since correcting old quality issues is outside the scope of this PR.

jamuhl:
@tisoap looks good, but I'm not a big fan of introducing breaking changes without having some compatibility options to get the old behavior. Not having those options we might force developers to keep the old version for a long time...and updating existing translations can be a lot of pain - (more pain than in other libs where you just have to update the code!).

If I see it right - all needed to do would be https://github.com/i18next/react-i18next/pull/981/files#diff-f518b658967844a5cbcb35ac5041c770R24 setting `children` to the value of `components` and set components to `null` - right?

Then an additional prop or global setting could enable the "legacy" behavior:

```js
// prop
<Trans defaults={} values={} components={} useLegacyComponents={true} />

// init
i18next.init({ react: { useLegacyComponents: true /* default false */ } })
```

This way we could also keep using the icu.macro without bigger changes: https://github.com/i18next/react-i18next/blob/master/icu.macro.js#L112 guess this builds nested components (but did not touch that code for long...so I'm not to sure, also there seems to be no nested test -> https://github.com/i18next/react-i18next/blob/master/test/__snapshots__/icu.macro.spec.js.snap , but adding the useLegacyComponents there at least would not break things for now)

tisoap:
I understand, I'll refactor to include flags for the old behavior once I have the time.

iamandrewluca:
Wouldn't this https://github.com/i18next/react-i18next/pull/859 PR be a better solution?

jamuhl:
@iamandrewluca would be...but that PR is pending for a long time...not sure what currently blocking it's completion...

Issue #915, opened 2019-08-08
AMD not working in IE11


roti:
**Describe the bug**
The AMD dist does not work in IE11 due to usage of the class syntax for `ReportNamespaces`.

**Occurs in react-i18next version**
10.11.5

**To Reproduce**
Load react-i18next in IE11.

**Expected behaviour**
IE11 should be supported.

**OS (please complete the following information):**
- OS: Windows 10
- Browser: IE 11.3024.14393.0

**Additional context**
AMD and UMD modules are build using `rollup.config.js`, where `@babel/preset-env` is configured with `targets.esmodules = true`. CommonJS seems to be fine, but I could not test it.

I'm not familiar with the effects of this setting, but apparently this causes babel not to target ES2015. Removing `targets.esmodules` had the expected effect for me, but I'm not 100% sure whether that's the correct solution. Here's my config from `rollup.config.js`.
```
const babelOptions = {
  presets: [
    '@babel/preset-env',
    '@babel/react',
  ],
  babelrc: false,
};
```

jamuhl:
hm...will have to look into this when I got more time - honestly not too familiar with that myself...

StevenDevooght:
Hi, we also ran into this issue.

jamuhl:
My guess preset-env per default picks browsers not older than 2 versions -> not sure about IE11 (but I would consider IE11 a dead horse => outdated browser)

roti:
Well, it's the default browser on Windows 10. :)

mie., 30 oct. 2019, 11:35 Jan Mühlemann <notifications@github.com> a scris:

lekoaf:
> Well, it's the default browser on Windows 10. :)

That's debatable. I'd argue that Edge is the default browser on Windows 10.

https://en.wikipedia.org/wiki/Microsoft_Edge#Features
> Microsoft Edge is the default web browser on Windows 10, Windows 10 Mobile, and Xbox One consoles, replacing Internet Explorer 11 and Internet Explorer Mobile.

roti:
Hmm. I have Windows 10 Version 1809 (OS Build 17763.805) and it came with IE 11.805.1773.0.

lekoaf:
> Hmm. I have Windows 10 Version 1809 (OS Build 17763.805) and it came with IE 11.805.1773.0.

Unfortunately Windows 10 comes with both IE 11 and Edge. However Edge is the default and recommended browser.

jamuhl:
[...]

PR #988, opened 2019-11-06
Update `rollup.config.js` for IE11 Transpilations

baublet:
Rollup doesn't set targets for browsers.

This commit addresses: https://github.com/i18next/react-i18next/issues/915

My question: is this an appropriate solution? My knowledge of the various JS module types isn't up to the point where I can say definitively one way or the other.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -15,6 +15,7 @@ const babelOptions = {
       {
         targets: {
           esmodules: true,
+          ie: "11"
         },
       },
     ],


baublet:
I tested this locally with IE11 and can confirm that targeting IE11 here _does work_, specifically because it transpiles ES6 classes into ES5.

jamuhl:
published in react-i18next@11.1.0

PR #992, opened 2019-11-12
Fix withTranslation eating up forwardedRef prop

### LINKED ISSUE #990: Updating to v11 can break prop forwarding when using withTranslation
**Describe the bug**
Forwarding a ref named `forwadedRef` (which is the [name react use in the examples in their docs](https://reactjs.org/docs/forwarding-refs.html#forwarding-refs-in-higher-order-components)) does not work on v11, but used to on v10.

I am not familiar with the best practices of forwarding refs with HOCs, and I can just rename this ref to something else to make our code work again, but it feels confusing that this specific name wouldn't work. If this behavior is indeed common and the maintainers don't feel like it's something worth fixing, this should probably at least be mentioned in the breaking changes of v11.

Seems to have been introduced by this commit since it's grabbing `forwardedRef` out of props: https://github.com/i18next/react-i18next/commit/299f4ff31339a0ace8d05e79b5e45e0a5bfe525e

**To Reproduce**
This code used to appropriately pass `ref` to the wrapped component in v10, it doesn't in v11.

```js
const InputWithRef = compose(
    withTranslation(),
    withProps(({ forwardedRef }) => ({ ref: forwardedRef }))
)(Input);

export const Input = React.forwardRef(
    (props, ref) => <InputWithRef {...props} forwardedRef={ref} />
);
```

**Expected behaviour**
For `withTranslation` to not remove `forwardedRef` from the props passed to its children.

**Screenshots**
N/A

- Device: N/A
- Browser: N/A

**Additional context**

emroussel:
When wrapping components with only the withTranslation HOC, this code works well:

```js
export const Input =  withTranslation('translation', { withRef: true })(Input),
```

But when using multiple ones, `withTranslation` was not propagating `forwardedRef` to its children, making code like this break (it also used to work in previous versions):

```js
const InputWithRef = compose(
    withTranslation(),
    withSomeOtherHOC(),
    withProps(({ forwardedRef }) => ({ ref: forwardedRef }))
)(Input);

export const Input = React.forwardRef(
    (props, ref) => <InputWithRef {...props} forwardedRef={ref} />
);
```

Fixes #990

### DIFF
--- a/src/withTranslation.js
+++ b/src/withTranslation.js
@@ -9,6 +9,7 @@ export function withTranslation(ns, options = {}) {
 
       const passDownProps = {
         ...rest,
+        forwardedRef,
         t,
         i18n,
         tReady: ready,


jamuhl:
published in react-i18next@11.2.0

felixmosh:
Hi, thank you for this PR.

I've updated to latest, and looks like now there is a prop (`forwardedRef`) that is passed to `Link` component (of Next), and that throws warning:
```
Warning: Failed prop type: Link: unknown props found: forwardedRef
    in Link
    in Link
    in withI18nextTranslation(Link)
    in withnextI18NextInternals(withI18nextTranslation(Link)) (at TextLink.tsx:14)
    in TextLink (at Logo.tsx:8)
```

adrai:
Perhaps only pass when options.withRef?

jamuhl:
best will be only adding it if it's not undefined

jamuhl:
fix was published in react-i18next@11.2.2

PR #999, opened 2019-11-18
Only pass forwardedRef to children if options.withRef is false

emroussel:
`forwardedRef` was always passed down to its children in #992, but as pointed in the comments, it should only be passed down if `options.withRef` is false (if it is true, we want to pass the ref as `ref` and remove `forwardedRef` from props) and if `forwardedRef` is defined.


jamuhl:
thank you - this fix was published in react-i18next@11.2.2

PR #1008, opened 2019-11-27
Store should be initialized after useSSR

fromi:
If a the store is initialized with data from Server side rendering, it fire these warnings anyway:
`i18next was not initialized... This means something IS WRONG in your application setup.`
However, the application setup is fine, the warning is inconsistent.
I think this PR will do the trick: i18n should be considered as initialize once it receive initialI18nStore datas.


jamuhl:
seems correct -> published in react-i18next@11.2.3

PR #1010, opened 2019-11-29
fixed graceful fallback for keys

ywongau:
t accepts an array of keys and will use first resolved key or fallback to the last key https://www.i18next.com/overview/api#t
But when i18next instance is not ready useTranslation hook simply returns an identity function, t(['doh', 'Human friendly fallback']) gives me ['doh', 'Human friendly fallback'] instead of 'Human friendly fallback'

### DIFF
--- a/src/useTranslation.js
+++ b/src/useTranslation.js
@@ -19,8 +19,9 @@ export function useTranslation(ns, props = {}) {
   if (i18n && !i18n.reportNamespaces) i18n.reportNamespaces = new ReportNamespaces();
   if (!i18n) {
     warnOnce('You will need pass in an i18next instance by using initReactI18next');
-    const retNotReady = [k => k, {}, false];
-    retNotReady.t = k => k;
+    const notReadyT = k => (Array.isArray(k) ? k[k.length - 1] : k);
+    const retNotReady = [notReadyT, {}, false];
+    retNotReady.t = notReadyT;
     retNotReady.i18n = {};
     retNotReady.ready = false;
     return retNotReady;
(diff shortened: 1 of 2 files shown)


jamuhl:
Instead of patching around getting a fallback from an array - shouldn't you fix the problem of accessing `t` before passing down `i18n`?!?

ywongau:
> Instead of patching around getting a fallback from an array - shouldn't you fix the problem of accessing `t` before passing down `i18n`?!?

We are using a dictionary whose value is [key, defaultMessage] for default language and main via she for others, which is similar to Message Descriptor in reacting
There are many benefits: Text for default language never fails; key being passed to t can be strongly t

jamuhl:
just your fix is https://github.com/i18next/react-i18next/pull/1010/files#diff-d5c6a91c6c8060c38a8b3541ef38332fR20

if (!i18n) ---> this is a mistake in your app that needs a fix ... not a patch

either set i18n via `I18nextProvider` or use `initReactI18next`

ywongau:
ok will close

jamuhl:
@ywongau look we can merge this - does not hurt - just configuring your application correctly you should never get into that code section where `!i18n` -> so the question stays how your code called `useTranslation|withTranslation` without having i18n passed yet by either `I18nextProvider` or use `initReactI18next`

ywongau:
Thanks, now reopened :)
Of course in my application react-i18next has to be configured correctly otherwise it wouldn't work
Having the message in default language as the last item in the key array and this patch is just to allow us to start using t to handle messages as soon as possible. Messages for default language will just work without any setup upfront

jamuhl:
published in react-i18next@11.2.5

If you like this module don’t forget to star this repo. Make a tweet, share the word or have a look at our https://locize.com to support the devs of this project.

If you liked my support / work - I would enjoy a coffee sponsored using the “:purple_heart:Sponsor Button”.

There are many ways to help this project :pray:

PR #1019, opened 2019-12-11
Allow html props on Trans, fix `parent` prop type

### LINKED ISSUE #1011: [Typescript] Trans component's additional props
**Describe the bug**
When Trans has a parent component (either because it's set as a prop, or because the configuration has one by default), it forwards its additional props (the one it does not consume) to that object. However in Typescript, this ends up with an error like:
```
error TS2322: Type '{ children: string; style: CSSProperties; parent: string; }' is not assignable to type 'IntrinsicAttributes & TransProps'.
  Property 'style' does not exist on type 'IntrinsicAttributes & TransProps'.
```

11.2.5

**To Reproduce**

Create a simple react-i18next Typescript strictly typed app and use a component:
```
<Trans parent="div" style={{color: 'green'}}>
  Hello world, React is super green
</Trans>
```

This works as I expect it to work at runtime (it creates a div, and style the text in green). However Typescript Compiler chokes on it with:
```
error TS2322: Type '{ children: string; style: CSSProperties; parent: string; }' is not assignable to type 'IntrinsicAttributes & TransProps'.
  Property 'style' does not exist on type 'IntrinsicAttributes & TransProps'.
```

**Expected behaviour**

I'd expect to accept the additional props happily. As a first step I would be fine if Trans was not checking those additional props, it's even better if it manages to check their types properly but I doubt it's doable especially if the parent is set in the config.

- OS: Linux 1adf3346df5d 5.3.0-19-generic #20-Ubuntu SMP Fri Oct 18 09:04:39 UTC 2019 x86_64 GNU/Linux
- Typescript: Version 3.7.2

rosskevin:
Closes #1011

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -29,7 +29,9 @@ declare module 'i18next' {
   }
 }
 
-export interface TransProps extends Partial<WithT> {
+export interface TransProps<E extends Element = HTMLDivElement>
+  extends React.HTMLProps<E>,
+    Partial<WithT> {
   children?: React.ReactNode;
   components?: React.ReactNode[];
   count?: number;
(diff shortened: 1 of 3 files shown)


rosskevin:
This can be a patch release.

jamuhl:
published in react-i18next@11.2.6

PR #1021, opened 2019-12-13
Allow ComponentType for Trans' `parent` type.

pcorpet:
I've also allowed `null` as this is the cleanest way I use to override the default parent:
* `undefined` -> use default parent from settings
* `string | React.ComponentType<any>` -> use in `React.createElement`
* `null` -> do not add a parent


pcorpet:
> Please update the usage tests in test/typescript and add these additional cases to prevent regressions.

Done. Thanks for the quick review.

rosskevin:
This can be a patch release

jamuhl:
sorry for the delay...was published in react-i18next@11.2.7

PR #1034, opened 2020-01-07
:memo: Update warning about react-native

Okazari:
I think the actual warning doesn't make sense anymore since the current versions of react-native support hooks ;) ! However, the warning is still usefull to keep in mind that you need 0.59 or higher :)


jamuhl:
thank you...good catch...

PR #1036, opened 2020-01-08
Fix the type of the components props of Trans.

pcorpet:


rosskevin:
@pcorpet this is certainly acceptable, but the `readonly` modifier only affects the implementation.  Since this is just a declaration without an implementation, I'm curious.   Why do you think this might be a useful addition?

pcorpet:
`readonly React.ReactNode[]` actually makes the array readonly not the react node. This is an indication that the `Trans` component has no intention of modifying the array (appending, sorting). This is useful as I have a readonly array myself that I want to set as components prop.

```jsx
const homeLinkComponents = [<a href="/" />] as const

<Trans components={homeLinkComponents} i18nKey="Go <0>home</0>" />
```

That currently fails with:

```
error TS4104: The type 'readonly [Element]' is 'readonly' and cannot be assigned to the mutable type 'ReactNode[]'.
```

rosskevin:
Ok, makes sense.  I would like you to add a usage test to the `test/typescript` dir and push it here so that we have no regressions.

Thank you!

rosskevin:

jamuhl:
published in react-i18next@11.3.0

PR #1045, opened 2020-01-15
fix useTranslation hook, stop breaking "Rules of Hooks", add checks for eslint.

Bajdzis:
The library is currently breaking the rules described in the React documentation.

https://reactjs.org/docs/hooks-rules.html

"Rules of Hooks" says :

> Don’t call Hooks inside loops, conditions, or nested functions. Instead, always use Hooks at the top level of your React function. By following this rule, you ensure that Hooks are called in the same order each time a component renders.

I fixed that and add checks "Rules of Hooks" in eslint.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,5 +1,6 @@
 {
   "extends": ["airbnb", "prettier", "prettier/react"],
+  "plugins": ["react-hooks"],
   "parser": "babel-eslint",
   "parserOptions": {
     "ecmaVersion": 8,
@@ -94,6 +95,7 @@
         "aspects": ["invalidHref"]
       }
     ],
-    "react/jsx-props-no-spreading": 0
+    "react/jsx-props-no-spreading": 0,
+    "react-hooks/rules-of-hooks": 2
   }
 }
--- a/package.json
+++ b/package.json
@@ -57,6 +57,7 @@
     "eslint-plugin-import": "2.18.2",
     "eslint-plugin-jsx-a11y": "6.2.3",
     "eslint-plugin-react": "7.16.0",
+    "eslint-plugin-react-hooks": "^2.3.0",
     "husky": "^3.0.3",
     "i18next": "^19.0.0",
     "jest": "24.8.0",
--- a/src/useTranslation.js
+++ b/src/useTranslation.js
@@ -8,6 +8,8 @@ import {
 } from './context';
 import { warnOnce, loadNamespaces, hasLoadedNamespace } from './utils';
 
+const notReadyT = k => (Array.isArray(k) ? k[k.length - 1] : k);
+
 export function useTranslation(ns, props = {}) {
   // assert we have the needed i18nInstance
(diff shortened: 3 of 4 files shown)


jamuhl:
Is there an issue...? The rule is non-conditional as it's always the same condition...

jamuhl:
https://github.com/i18next/react-i18next/commit/03553c40b4965985b344b9adcaff4405e805752b

makes it equal to the usage in useTranslation - still if the rule of hooks get broken in those cases - there is a bigger issue in your app that needs attention!!!

Bajdzis:
> Is there an issue...? 

My application, throw critical errors when components re-rendering. I must use lazy start `i18next.init()` but when call `init` function after first render my app is crash. 

> The rule is non-conditional as it's always the same condition...

Not in my case :/  All I did was move hooks before return instruction. This return is a problem because occurs after useState and useEffect.

https://github.com/i18next/react-i18next/blob/master/src/useTranslation.js#L27

jamuhl:
The rule of react-i18next is `You shall pass an i18n instance - initialized or not - or we might throw an error here in future` https://github.com/i18next/react-i18next/blob/master/src/useTranslation.js#L20

Bajdzis:
i use midelware initReactI18next
```
import { initReactI18next } from 'react-i18next';
```
midelweare in i18n run when call init. How pass i18n without init?

jamuhl:
call changeLanguage instead of init to set a language...

Bajdzis:
@jamuhl  yes... awesome idea! Download two language (first for fix crash, second for user) because "react-i18next" package have **BUG**. 

You can download more data and make inefficient applications! 

Why don't you want to fix this **BUG**?

jamuhl:
No need to get rude...you can use a language detector for language detection, even custom, even async...there is no need to delay the init call...

adrai:
There is probably a misunderstanding here.
`i18next.init` should only be called once (this is how this library is designed)
I would also recommend throwing an error [here](https://github.com/i18next/react-i18next/blob/master/src/useTranslation.js#L20) instead of a warning.
I'm in favor of using the library "correctly" and not allow such strange constellations.
Regarding `changeLanguage`, yes the correct way to do this is with a language detector: https://www.i18next.com/misc/creating-own-plugins#languagedetector

Bajdzis:
@jamuhl 
I'm sorry, I think I may have overreacted back there. But i don`t understand why you don't write correct hooks and breaking ReactJS documentation. 😢 

I have now found a workaround by using setI18n, but @adrai  solution looks much better. Thanks for help!
```
import { setI18n } from 'react-i18next';

const instancei18n = i18n;

setI18n(instancei18n);

const runLang = () => instancei18n.init(....)
```

Still, I think you should write this package with good practices ReactJs docs.

jamuhl:
We do write this package with good practices - again there is no rule broken if following the rules of react-i18next which is there needs to be an i18n instance... but won't discuss this over and over again...

your code is "fixed" - the rule no longer broken...

PR #1047, opened 2020-01-27
fix(Trans): add missing `additionalProps` type

iamandrewluca:
https://github.com/i18next/react-i18next/blob/master/src/Trans.js#L216
https://www.typescriptlang.org/docs/handbook/advanced-types.html#index-types-and-index-signatures

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -43,6 +43,7 @@ export interface TransProps<E extends Element = HTMLDivElement>
   tOptions?: {};
   values?: {};
   t?: TFunction;
+  [additionalProps: string]: any;
 }
 export function Trans<E extends Element = HTMLDivElement>(props: TransProps<E>): React.ReactElement;
 


iamandrewluca:
I agree with you. I'll close this for now.
Solution would be to use `Trans` this way

```jsx
<Trans<any> i18nKey="theKey" parent="p" className="bg-light">
    Hellow World!
</Trans>
```

iamandrewluca:
Looked at `Typography` component and narrowed it down to something like this

image


```jsx
import React, { ComponentPropsWithoutRef } from 'react';

interface TransProps<E extends React.ElementType> {
  i18nKey?: string;
  parent?: E;
  // ...
}

function Trans<E extends React.ElementType = 'div'>(
  props: TransProps<E> & Omit<ComponentPropsWithoutRef<E>, keyof TransProps<E>>
) {
  return null;
}

interface TestParentProps {
  someProps: boolean;
}
function TestParent(props: TestParentProps) {
  return <div />;
}

(
  <React.Fragment>
    <Trans parent="a" href="" />
    <Trans parent="label" htmlFor="some-id" />
    <Trans href="" className="" />
    <Trans parent="a" href />
    <Trans parent={TestParent} className="" href />
  </React.Fragment>
);
```

rosskevin:
That looks good. The only critique I would make is to also default the div on the TransProps so people that use the interface to composite others get it e.g. `TransProps<E extends React.ElementType = 'div'>`

iamandrewluca:
Actually I found another solution, that does not default on anything, because parent is optional.

Also I am using `React.ElementType` instead of `Element` because `Element` is not a component type, but `React.ElementType` can be a html tag, function component or class component

rosskevin:
`parent` is an optional override, see https://github.com/i18next/react-i18next/blob/master/src/Trans.js#L231

It seems by default it is `undefined` as of 7.0.0 - see https://github.com/i18next/react-i18next/blob/c764b0246e20844beaa5c01cc1d6af2ea0a4b347/CHANGELOG.md#700

So, technically, if you do not setup `defaultTransParent` in the init options, then https://github.com/i18next/react-i18next/blob/master/src/Trans.js#L255 will not use the `additionalProps`, unless you explicitly pass in the `parent` prop.

So, I think you'll have to experiment with `Trans<E extends React.ElementType = undefined>` or something of that nature, because we want to accommodate default behavior first.  Unfortunately I don't see a way this is going to just work for both default and init overrides because we can't inspect the init at runtime.  Perhaps there are some tricks to make this more amenable.

PR #1050, opened 2020-02-09
Add optional defaultNS

### LINKED ISSUE #1049: Missing defaultNS prop in I18nextProviderProps
**Describe the bug**
The definition file `src/index.d.ts` seems to be outdated

**11.3.1**

**To Reproduce**

1. Create a new dummy `.tsx` file with vscode
2. Include a snippet like this one:
```ts
import { I18nextProvider, withTranslation } from 'react-i18next'

<I18nextProvider i18n={i18n} defaultNS={screenId}>
</I18nextProvider>
```

**Expected behaviour**
`defaultNS` should not show the error

**Screenshots**
If applicable, add screenshots or a GIF to help explain your problem.
image

- Device: N/A
- Browser: N/A

jpgarcia:
Closes #1049

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -98,6 +98,7 @@ export function withTranslation(
 
 export interface I18nextProviderProps {
   i18n: i18n;
+  defaultNS: string;
 }
 
 export const I18nextProvider: React.FunctionComponent<I18nextProviderProps>;
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -98,7 +98,7 @@ export function withTranslation(
 
 export interface I18nextProviderProps {
   i18n: i18n;
-  defaultNS: string;
+  defaultNS?: string;
 }
 
 export const I18nextProvider: React.FunctionComponent<I18nextProviderProps>;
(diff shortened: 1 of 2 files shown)


jamuhl:
Will be able to publish this after my holidays (this weekend, early next week)

test/typescript/I18nextProvider.test.tsx
```diff
@@ -23,7 +23,7 @@ i18next.use(initReactI18next).init({
 function test() {
   return (
     <React.Suspense fallback={<p>Loading</p>}>
-      <I18nextProvider i18n={i18next}>
+      <I18nextProvider i18n={i18next} defaultNS={'translations'}>
```
rosskevin: Please don't change the current test because it is a valid case on it's own.  Please add a new `function testDefaultNS()`

src/index.d.ts
```diff
@@ -98,6 +98,7 @@ export function withTranslation(
 
 export interface I18nextProviderProps {
   i18n: i18n;
+  defaultNS: string;
```
rosskevin: This is not a required property, only `i18n` is.  This should be `defaultNS?: string`.  Leaving the test as-is and adding another would prove this true in the future.

rosskevin:
Thanks @jpgarcia 

jamuhl:
was published in react-i18next@11.3.2

PR #1051, opened 2020-02-10
[useTranslation] Avoid setting the new `t` function if the component is unmounted.

zozoens31:
Once the translations are loaded from backend, the hook sets its `t` function without checking that the component is still mounted, which induces an error from `react-dom`.

```Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.```

I've updated the code to keep the `isMounted` pattern to be consistent with the previous version, but I'll gladly change it for a promise cancellation if you find it more appropriate.

### DIFF
--- a/src/useTranslation.js
+++ b/src/useTranslation.js
@@ -1,4 +1,4 @@
-import { useState, useEffect, useContext } from 'react';
+import { useState, useEffect, useContext, useRef } from 'react';
 import {
   getI18n,
   getDefaults,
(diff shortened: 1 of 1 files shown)


sprilukin:
+1 for the issue.

jamuhl:
Looks valid...never run into this myself - but thinking about it the case is valid. Will deeper check this when back from holidays (this weekend, early next week) and merge this into master.

jamuhl:
sorry forgot about this one...published in react-i18next@11.3.4

PR #1092, opened 2020-04-18
fix returning defaultValue for Trans component

WebKieth:


jamuhl:
thank you ... was published in react-i18next@11.3.5

PR #1097, opened 2020-04-23
Add sideEffects false to package json to allow tree shaking

kenny-f:
Noticed that all the exports were being included output bundles even though only a few were used.

This pr adds `sideEffects: false` to the `package.json` so that bundlers like webpack can tree shaking unused code. 

BEFORE:
Screenshot 2020-04-23 at 09 01 17

AFTER:
Screenshot 2020-04-23 at 09 21 17

### DIFF
--- a/package.json
+++ b/package.json
@@ -147,5 +147,6 @@
       "**/react-i18next.min.js"
     ]
   },
-  "lock": false
+  "lock": false,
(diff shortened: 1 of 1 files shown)


jamuhl:
sorry took a little longer. published in react-i18next@11.4.0

PR #1098, opened 2020-04-28
refactor(context): remove redundant check `hasUsedI18nextProvider`

iamandrewluca:
There is no need for `getHasUsedI18nextProvider`
If `I18nContext.Provider` was used then `useContext(I18nContext`) will return a value.
In cases when user overrided `I18nContext.Provider` value we fallback to an empty object `useContext(I18nContext) || {}`


jamuhl:
will be released with next version

PR #1099, opened 2020-04-28
refactor(Trans): add note for `parent !== undefined` check

iamandrewluca:


jamuhl:
Thank you...will be published with next release.

PR #1102, opened 2020-04-29
refactor(nodesToString): flatten logic

iamandrewluca:


iamandrewluca:
While refactoring this I encounntered this case that must be a bug I think.

e.g. `lorem {{value}} ipsum`

For example user changes default interpolatiion prefix and suffix

```
i18next.init({
    interpolation: {
		prefix: '{',
		suffix: '}'
	}
});
```

Shouldn't node to string return e.g. `lorem {value} ipsum`?

jamuhl:
regarding interpolation - yes that was hardcoded to use `{{ ... }}`...honestly I'm at a point where I'm willing to remove setting those prefix/suffix in future...(it makes it just harder to support i18next in translation management software by not defining a real standard but letting most open to configuration)

iamandrewluca:
@jamuhl what issue did you encountered with interpolation prefix/suffix? Whay do you mean by:

> it makes it just harder to support i18next in translation management software

ps: Many users may still use interpolation prefix/suffix for whatever reasons.

jamuhl:
There is no issue - I just did not add using the prefix/suffix from i18next.options (out of laziness or just forgot)

Translation Management Software often use some sort of xliff or similar to convert syntax of an i18n lib to some form reusable -> so eg. interpolation values can be replaced with placeables, or just highlight the syntax ---> having a syntax without fixed {{}} usage makes it more or less impossible to correctly parse those i18next json in cases devs override the defaults.

iamandrewluca:
I got it. It makes translators life harder :smile:

jamuhl:
Sure...should be changed to support it...

Issue #1117, opened 2020-05-19
Support passing of attributes from translation i.e. <0 href="web address"></0> to react component i.e. <a></a>


chaosmaker:
I am currently building a webpage with multi language support with extensive language files and numerous inline links that I like to add with <Trans> component. In some cases, I have multiple links within a single <Trans> component. I think for readability it would be great if I could just add the attribute to a given <0></0> replacement component. 

For example take the following translation file:
```
"privacy-test": {
  "title": "Lorem ipsum dolor sit amet, consectetur adipiscing elit",
  "text": "Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat",
  "list": [
     {
	"_id": 1,
	"title": "Curabitur pretium tincidunt lacus",
	"sublist": [
	  {
	    "_id": "1.1",
	    "title": "Sed ut perspiciatis, unde omnis iste natus error",
	    "text": "totam rem aperiam eaque ipsa, quae ab illo inventore veritatis et quasi architecto beatae<0></0>vitae dicta sunt, explicabo<0></0>Nemo enim ipsam voluptatem, quia voluptas sit<1 href=\"www.google.com\">www.google.com</1>"
	  }
      ]
    }
  ]
}
```
With a <Trans> component as follows:
```
<Trans ...>
  <br/><a></a>
</Trans>
```
which I would like to result in:
```
<h1>Sed ut perspiciatis, unde omnis iste natus error</h1>
<p>totam rem aperiam eaque ipsa, quae ab illo inventore veritatis et quasi architecto beatae</br>vitae dicta sunt, explicabo</br>Nemo enim ipsam voluptatem, quia voluptas sit<a href=\"www.google.com\">www.google.com</a>.</p>
```
As I am using a loop I don't want to use the more standard way of having the text within the Trans component but just want to add the <0></0> counters into my many language files. 
I don't think the passing of attributes is currently possible, or rather I haven't figured it out. I don't think it would be too difficult to implement. Would it be possible to have such a functionality?

adrai:
I don’t know exactly what you want to do...
But here is an example with attribute...
just use t() https://react.i18next.com/latest/trans-component#using-with-react-components

chaosmaker:
Well that is exactly what I don't want to do. While I have those version in my files as well, they require me to specify a) how many t()s I will be adding and then I need to implement them as well. What I am after is if I have the following code:
```
list.map((title,text) => {
return(
[...]

PR #1120, opened 2020-05-20
Added lodash and merge option in mapAST

chaosmaker:
Dear maintainers of react-i18next,

Below is my pull request for #1117 where I initially requested the option to have props within a `<0 ...props></0>` passed to its final component. It is a very simple fix. 

I have tried running `npm run test` but it seems the distribution lacks `dtslint/dtslint.json` - I have tested the code extensively and don't see any issues that may occur. 

I also wasn't sure where to change the documentation. If you point me in the right direction I will adapt where necessary. 

#### Checklist

- [ ] tests are included

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,8 +25,7 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.3.1",
-    "html-parse-stringify2": "2.0.1",
-    "lodash.merge": "^4.6.2"
+    "html-parse-stringify2": "2.0.1"
   },
   "devDependencies": {
     "@babel/cli": "^7.2.3",
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
   "dependencies": {
     "@babel/runtime": "^7.3.1",
     "html-parse-stringify2": "2.0.1",
-    "lodash": "^4.17.15"
+    "lodash.merge": "^4.6.2"
   },
   "devDependencies": {
     "@babel/cli": "^7.2.3",
--- a/package.json
+++ b/package.json
@@ -25,7 +25,8 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.3.1",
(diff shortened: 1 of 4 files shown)


adrai:
Is full lodash dependency really needed?

Can't the same be achieved also with: `const child = Object.keys(node.attrs).length !== 0 ? { props: node.attrs, ...tmp } : tmp;` ?

chaosmaker:
I tried that first and couldn't get it to work, the props are overwritten by `{}` that is in tmp. _.merge has the advantage that it will deep merge the properties, so if there are defined properties like class etc, they will be merged together as opposed to replaced. To reduce full lodash I have changed it to load lodash.merge only.

jamuhl:
sadly won't accept any lodash dependency...this adds over 3.7kb gzipped: https://bundlephobia.com/result?p=lodash.merge@4.6.2

also please add some test cases for future regression:

- asserting feature works
- not overwrites props set in code (props already on the node)

thank you 👍

chaosmaker:
Ok I have removed lodash and just focused on merging the props using a small function. So nice and light. The code is set in such a way that original values in `<Trans><anything ...prop></Trans>` are not overwritten when set in the language file. Only missing props are added. 

Also not sure why coverage decreased? I didn't remove anything?

I am a bit new to github/code reviews - how/where do I add some test cases?
> also please add some test cases for future regression:
> 
> * asserting feature works
> * not overwrites props set in code (props already on the node)

jamuhl:
test can be added here: https://github.com/i18next/react-i18next/blob/master/test/trans.render.spec.js

coverage decreased as your newly added code is not covered.

chaosmaker:
I added some tests which seem to work. Hope that is all you need!

jamuhl:
looks ok...will check this in detail tomorrow (merge & publish)

adrai:
Thanks, it's included in v11.5.0

PR #1128, opened 2020-05-28
Fix typo in withTranslation() example.

herpiko:
`const { t } = this.props` leads to error. `t` should be derived from `this.props.t`.

No unit test or documentation added since this is in example code.

#### Checklist

- [ V ] only relevant code is changed (make a diff before you submit the PR)
- [ V ] run tests `npm run test`
- [ X ] tests are included
- [ X ] documentation is changed or added


adrai:
Changing this breaks it...
It is correct as is.

PR #1130, opened 2020-05-31
respect overloadTranslationOptionHandler if tOptions are not defined

adrai:
#### Checklist

- [ ] tests are included

shuld fix https://github.com/i18next/i18next-emoji-postprocessor/issues/1 when used globally


jamuhl:
https://www.i18next.com/overview/configuration-options#translation-defaults postProcess option is the correct solution

The overload is more the t(key, str) -> str = default (or other function overrides)

PR #1140, opened 2020-06-18
feat: Trans interpolating self-closing tags in components prop(object)

### LINKED ISSUE #1136: Trans alternative usage with Self-Closing Tags
## 🐛 Bug Report

I'm not sure if intended, but using the alternative usage of `Trans`, as described in [docs](https://react.i18next.com/latest/trans-component#alternative-usage-v-11-6-0), doesn't seem to support interpolating tags with no content. This new feature is really game-changing, so I believe we should support it even without passing content to it!

Also, the type definitions also don't seem to match the new possible `components` structure, as it's accepting only `readonly React.ReactNode[] | undefined`.

## To Reproduce

[Link to CodeSandbox](https://codesandbox.io/s/i18next-u7lez)

## Expected behavior
Trans component interpolation regardless of tags having content inside them or not.

## Your Environment

- *runtime version*: Firefox
- *i18next version*:
  - *i18next*:  **19.4.5**
   - *react-i18next*: **11.6.0**
- *os*: Linux

abbudao:
This PR intends to close #1136. It provides:
1. More versatility to the Trans component as now is possible to interpolate self-closing tags
2. Updated Trans typing, reflecting the alternative usage described in [docs](https://react.i18next.com/latest/trans-component#alternative-usage-v-11-6-0).
3. Improved readability on `mapAst`'s reducer.

The third point is opinionated, but I was having trouble to reason with it the way it was written. 
I was also forced to write the `pushTranslatedJSX` function (still unsure if this is the best name though) to keep the DRY principle. 

#### Checklist

- [X] tests are included
- [X] documentation is changed or added

### DIFF
--- a/src/Trans.js
+++ b/src/Trans.js
@@ -135,6 +135,17 @@ function renderNodes(children, targetString, i18n, i18nOptions, combinedTOpts) {
   // -> avoids issues in parser removing prepending text nodes
   const ast = HTML.parse(`<0>${interpolatedString}</0>`);
 
+  function pushTranslatedJSX(child, node, rootReactNode, mem, i) {
+    const childs = getChildren(child);
+    const mappedChildren = mapAST(childs, node.children, rootReactNode);
+    // console.warn('INNER', node.name, node, child, childs, node.children, mappedChildren);
+    const inner =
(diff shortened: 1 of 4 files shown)


jamuhl:
thank you....was published in react-i18next@11.7.0

PR #1142, opened 2020-06-29
Fixed typescript module

sgregoire:
In the PR #945, the module was overriden, Typescript could not see any of the keys defined in `i18next.i18n`.

Only the following properties these two properties were accessible: `reportNamespaces`, `reportNamespaces`.

Since I could not find any documentation about these two methods, I've decided to remove the module override.

#### Checklist

- [X] tests are included
-  [ ] documentation is changed or added (NA)

### DIFF
--- a/test/typescript/context.test.tsx
+++ b/test/typescript/context.test.tsx
@@ -34,6 +34,3 @@ i18n.setDefaultNamespace;
 i18n.store;
 i18n.t;
 i18n.use;
-// To be reviewed: No reference found in official documentation.
-// i18next.reportNamespaces.addUsedNamespaces(['translation']);
-// i18next.reportNamespaces.getUsedNamespaces();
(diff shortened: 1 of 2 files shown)


test/typescript/context.test.tsx
```diff
@@ -1,4 +1,36 @@
-import i18next from 'i18next';
+import i18n from './i18n';
```
rosskevin, line 1: This import is incorrect.  You are importing a local file, instead of the module, which would disable module augmentation.  This should be switched back.

rosskevin:
@sgregoire it is unclear what problem you are solving.  Module augmentation works, so there is no need to remove the augmentation added.

We can accept your additional usage tests to see the keys exist, but the import needs to be changed back and the augmentation and tests should remain.

PR #1160, opened 2020-08-18
fix: typescript definition of context object

seanplwong:
Wrap I18nContext value in an object

FIX: https://github.com/isaachinman/next-i18next/pull/799
FIX: https://github.com/isaachinman/next-i18next/issues/789

#### Checklist

- [x] tests are included


seanplwong:
In fact I could help get some types generated at build time later, so that manual maintenance on types could be reduced

rosskevin:
Thanks @seanplwong, @jamuhl this can be a patch release

jamuhl:
published in react-i18next@11.7.1

Issue #1124, opened 2020-05-25
Issue with React 16.13: Cannot update a component [ X] while rendering a different component [Y].


ifmael:
## 🐛 Bug Report
React show a warning when a function component is updated during another component's render phase.  https://fb.me/setstate-in-render

## To Reproduce

The  [codesandbox](https://codesandbox.io/s/github/ifmael/ifmael-web)  to reproduce the issue:
- Open console.
- Click on "en"

## Expected behavior
I don't expect to see a warning in the console

## Your Environment

- *runtime version*: 12.12
- *i18next version*: >= 11.5.0
- *os*:  Windows 10

jamuhl:
Looks like more related to your sample than to what react-i18next provides?

ifmael:
If I remove `  const { t } = useTranslation();`  from my components and use normal string , it doesn't happen. 
In this message https://github.com/facebook/react/issues/18178#issuecomment-595846312 , gaeron shows how to identify the source of the issue.

![i18n-issue](https://user-images.githubusercontent.com/217620/82822264-66349e80-9e9d-11ea-844f-965a5e815fa2.PNG)

In my component `locale-provier.js`, I only change the locale based on the page context

```javascript
import React from 'react'
import LocaleContext from './locale-context'
import { useTranslation } from "react-i18next"

const LocaleProvider = ({ children, locale}) =>{
  const { i18n } = useTranslation()

  i18n.language !== locale && i18n.changeLanguage(locale);

  return <LocaleContext.Provider value={locale}>{ children }</LocaleContext.Provider>
}

export default LocaleProvider
```

jamuhl:
Still, the problem does not exist in our samples...so it's specific to your setup, configuration...

ifmael:
Ok. How can i identify where is the problem? Because I don't have idea what is it going  on

jamuhl:
try:

- adding a lng on init or using a language detector (at least in the sandbox you get a warning for that)
- set `saveMissing: false` on init (that might trigger a rerender - saveMissing should not be used in production anyway)
- try playing with `useSuspense: false` in react-i18next options: https://react.i18next.com/latest/i18next-instance

CrashyBang:
@ifmael any luck? I am seeing this to, is to do with `isMounted` being changed in the `useTranslation` hook during a render.

jamuhl:
[...]

PR #1165, opened 2020-08-26
Avoid setState while react is rendering

rmedaer:
As documented in [my comment about issue #1124](https://github.com/i18next/react-i18next/issues/1124#issuecomment-680153621), this fix should avoid `setState` while react is rendering.

See #1124 for more details.

#### Checklist

-  ~tests are included~
-  ~documentation is changed or added~

### DIFF
--- a/src/useTranslation.js
+++ b/src/useTranslation.js
@@ -85,7 +85,6 @@ export function useTranslation(ns, props = {}) {
   // not yet loaded namespaces -> load them -> and trigger suspense
   throw new Promise(resolve => {
     loadNamespaces(i18n, namespaces, () => {
-      if (isMounted.current) setT(getT());
       resolve();
     });
   });


rmedaer:
@jamuhl, I don't know how to fix the coverage issue, it's (I guess) a side effect which (according to the report) is not in `src/useTranslation.js`. If you really need me to fix the coverage for this patch, I would enjoy some help.

Kind regards,

jamuhl:
@rmedaer not needed...will check PR this asap...thank you

jamuhl:
published in react-i18next@11.7.2

Thank you for figuring this out and providing the PR. Hopefully this will not break any side cases - but looks safe in my tests.

rmedaer:
> published in react-i18next@11.7.2

Thank you.

> Hopefully this will not break any side cases

Let me know if you find related issue.

PR #1174, opened 2020-09-23
Avoid redundant re-rendering

takakobem:
Context value should be memoized to avoid re-rendering in consumer components.
Actually, when I18nProvider is re-rendered event if props are not changed, value object is recreated and all consumer components are re-rendered.

#### Checklist

- [x] tests are included


jamuhl:
Good catch...👍

jamuhl:
published in react-i18next@11.7.3

Issue #1209, opened 2020-12-04
Issue with the mock of the Trans component in the jest testing example


moT01:
## 🐛 Bug Report

Potential bug with the mock of the `Trans` component in [the jest example](https://github.com/i18next/react-i18next/blob/master/example/test-jest/src/__mocks__/react-i18next.js)

I came across [this issue](https://github.com/i18next/react-i18next/issues/927) after seeing a similar problem described there. I dug into it and the issue is that if there is only one child of the `<Trans>` component, then `children` would be a single object, which was getting passed to the `renderNodes` function - if there is more than one child, `children` is an array and the function would work as intended.

## To Reproduce

I was able to reproduce this by running the jest example, making the [Trans component](https://github.com/i18next/react-i18next/blob/master/example/test-jest/src/Trans.js#L9) look like this:
```js
<Trans t={t}>
  <strong>description.bold</strong>
</Trans>
```
and then running the tests. You get a `Cannot convert undefined or null to object` error.

My fix:
```js
const Trans = ({ children }) =>
  !children.length ? renderNodes([children]) : renderNodes(children);
```

I confirmed this fix by adding this code into the example mock and running the tests again. If this is a suitable fix, I can make a PR.

tigerabrodi:
I would try reproducing and have a look, aside from that, help is very welcome :two_hearts:

tigerabrodi:
Regarding your fix, to me, I wonder why we do not check if it is an array?

Like this 
```js
const Trans = ({ children }) =>
  Array.isArray(children) ? renderNodes(children) : renderNodes([children])
```

tigerabrodi:
I was able to reproduce your error and my fix is also working :tada:.

### Few Images
![Screenshot from 2020-12-04 23-33-26](https://user-images.githubusercontent.com/49603590/101222038-3acf0f80-3689-11eb-896e-71df763580d7.png)
![Screenshot from 2020-12-04 23-33-22](https://user-images.githubusercontent.com/49603590/101222048-3efb2d00-3689-11eb-8a32-75e148c62189.png)
![Screenshot from 2020-12-04 23-33-12](https://user-images.githubusercontent.com/49603590/101222055-428eb400-3689-11eb-9c64-780ca945b11c.png)

tigerabrodi:
@moT01 I'd say go for a PR :tada: :raised_hands:

PR #1210, opened 2020-12-04
Fix/trans mock in jest example

moT01:
#### Checklist

- [ ] tests are included

Closes: #1209


tigerabrodi:
Looks nice @moT01 :tada: :raised_hands: , also something I wanted to mention is, why `Array.isArray` is better for this case, is because when you check for length only, Maps and Sets do have a length property in JavaScript, therefore it can also be confusing for the reader of the code :+1:

adrai:
lgtm

jamuhl:
thanks to all for helping with this one....

moT01:
Thanks for the quick reviews and merge on this everyone 🎉 I'm having a lot of fun using this package by the way. It's real easy to use - nice job 👍

tigerabrodi:
@moT01 Happy to hear :smiling_face_with_three_hearts: :raised_hands:

Issue #1213, opened 2020-12-07
(typescript) Template literals no longer work post 11.8.0


onionhammer:
Example usage

```ts
const [ t ]      = useTranslation();

return <>{t`Hello`}</>
 
```

Worked in react-i18next v 11.7.4, but not 11.8.0, now errors with

```
No overload matches this call.
  Overload 1 of 2, '(key: never, options?: string | TOptions<StringMap> | undefined): never', gave the following error.
    Argument of type 'TemplateStringsArray' is not assignable to parameter of type 'never'.
  Overload 2 of 2, '(key: never, defaultValue?: string | undefined, options?: string | TOptions<StringMap> | undefined): never', gave the following error.
    Argument of type 'TemplateStringsArray' is not assignable to parameter of type 'never'.  TS2769
```

tigerabrodi:
Seems to be something that happened after TypeScript PR. Any ideas @pedrodurek ?

pedrodurek:
it is the same error https://github.com/i18next/react-i18next/issues/1212

tigerabrodi:
> it is the same error #1212

@pedrodurek I should defo debounce my questions sometimes haha :man_facepalming: :joy:

pedrodurek:
Hey @onionhammer, It should be fixed now. I added the support to template string back, but unfortunately `TemplateStringsArray` doesn't support generics yet (https://github.com/microsoft/TypeScript/issues/33304), so this PR https://github.com/i18next/react-i18next/pull/1193, which types all keys from `t` function won't work for you.

@tigerabrodi you were right! This was a different issue indeed, which was returning the same error.

pedrodurek:
Hey @onionhammer, can we close this one?

onionhammer:
@pedrodurek seems like this leads to another breaking change/or perhaps bug; now the result of t`Something` is not a string, it's a TFunction<string>, so it cant be passed as a string to props.

pedrodurek:
Ohh yeah, it's been handled here: https://github.com/i18next/react-i18next/issues/1217

onionhammer:
Closing in favor if #1217

Issue #1212, opened 2020-12-07
Linting of TFunction fails since the last update


mxsxs2:
## 💥 Regression Report

Linting of `TFunction` fails since the last update. The `t` function should accept its `key` parameter to be `string | string []`. However, it seems to expect to be `never`.

`K` seems to resolve to `never` at https://github.com/i18next/react-i18next/blob/a4aaf65e937e2b5ef021be41d581614a9aa80c73/src/ts4.1/index.d.ts#L91

## Last working version

Worked up to version: 11.7.4

Stopped working in version: 11.8.0

## To Reproduce

```js
import React from 'react'
import { useTranslation } from 'react-i18next/*'

const MyComponent= ()=>{
    const {t} = useTranslation()
    return <span>{t('hello')}</span>
}

export default MyComponent
```

## Expected behavior

I expect to not have _Argument of type 'string' is not assignable to parameter of type 'never'.ts(2769)_ at line 6 for `t('hello')`

Full error:

>  error TS2769: No overload matches this call.
  Overload 1 of 2, '(key: never, options?: string | TOptions<StringMap> | undefined): never', gave the following error.
    Argument of type 'string' is not assignable to parameter of type 'never'.
  Overload 2 of 2, '(key: never, defaultValue?: string | undefined, options?: string | TOptions<StringMap> | undefined): never', gave the following error.
    Argument of type 'string' is not assignable to parameter of type 'never'.
## Your Environment

- *runtime version*: node 14.15.1
- *i18next version*: 11.8.0
- *TypeScript version*: 4.1.2
- *os*: Windows with Ubuntu WSL2

paulrostorp:
same here:
![Screenshot 2020-12-07 at 14 52 58](https://user-images.githubusercontent.com/1866649/101362774-d5ee0200-38a0-11eb-9880-5dd25238f4b6.png)

adrai:
@pedrodurek can you have a look at this?

JeroenReumkens:
Same here, was looking into failing builds for the past 30 mins. This only seems to be the case in the 4.1 typings, 4.0.x is fine. At least it helped me find out our CI users a different TS version compared to my local environment 😅

pedrodurek:
sure

goldmont:
Same here.

pedrodurek:
[...]

PR #1214, opened 2020-12-08
fix: typescript definitions for t function without namespaces

pedrodurek:
Fix bugs: https://github.com/i18next/react-i18next/issues/1213 and https://github.com/i18next/react-i18next/issues/1212

In addition, I also added two examples of use and included the reference in the README file. Moreover, now that prettier is supporting typescript 4.1 (https://github.com/prettier/prettier/issues/9234) I also bumped the prettier version and run it on type definitions file.

#### Checklist

- [ ] tests are included
- [x] documentation is changed or added


Thanaen:
Hi !
I've tried to apply the new TS 4.1 example, but it seems to break when a there is a key named "keys" in the translations file :-(
Here is a simple repo I used to reproduce this issue. Simply clone it, do 'yarn' then 'yarn lint' / 'yarn start' !
Here is the repository: https://github.com/Thanaen/repro-react-i18n-ts-bug

jamuhl:
@Thanaen this looks like a different issue?!? My suggestion would be creating a new issue for that and mentioning @pedrodurek

Thanaen:
@jamuhl Okay, creating a new issue ! Thanks for your suggestion 👍

jamuhl:
was published in react-i18next@11.8.1

Issue #1217, opened 2020-12-08
Build has Error by Type Type 'string' is not assignable to type 'string[]', but enable returnObjects


Him-2C:
## 🐛 Bug Report

Has error on 11.8.1 or newer, but 11.7.3 it's work!

![Screen Shot 2563-12-09 at 00 56 31](https://user-images.githubusercontent.com/30730045/101522446-cf9f7900-39b9-11eb-8bbf-debd79aef711.png)

## Your Environment

- *runtime: node v14, nextjs
- *react-i18next version*: ^11.8.0
- *os*: Mac

adrai:
//cc @pedrodurek

pedrodurek:
Hey @Him-2C, I'll change the default TFunctionResult to accept other types. Right now, if you rely on the type augmentation as in this example here https://github.com/i18next/react-i18next/blob/master/example/react-typescript4.1/namespaces/%40types/react-i18next/index.d.ts, you're problem should be gone.

Him-2C:
@pedrodurek Thank, but I don't DefaultResources.

I set resource by this
![image](https://user-images.githubusercontent.com/30730045/101526913-e3e67480-39bf-11eb-8a76-01553f31f8e3.png)

pedrodurek:
Hey @Him-2C, it should be fixed now in v11.8.2.

Him-2C:
I have to try build again, but not work. It seems the new type with object.

![image](https://user-images.githubusercontent.com/30730045/101645516-0d0f0f80-3a69-11eb-8b88-b624c2c17d20.png)

john-d-pelingo:
Hello! o/

I am new to this and my question would be why is the `t` now returning an `object` as well? Shouldn't it just be a `string` or `null` like before?

BenJenkinson:
Hi @pedrodurek, 

The fix you made has introduced a new problem for me.

The following code does not work, and produces a type/build error.

```ts
const { t } = useTranslation("exampleNamespace");

const foo: string = t("exampleKey");
```

```
Type 'TFunctionResult' is not assignable to type 'string'.
  Type 'undefined' is not assignable to type 'string'.
```

I think the `TFunctionResult` type you're importing from `i18next` represents _all possible return types_, and not what it actually would be. The default return type is `TResult`, which must extend from `TFunctionResult`, but is defaulted to `string`.

[...]

PR #1221, opened 2020-12-09
Fix: Return type inference for t function (typescript 4.1)

pedrodurek:
Fix new issues reported here: https://github.com/i18next/react-i18next/issues/1217

#### Checklist

- [ ] tests are included


jamuhl:
published in react-i18next@11.8.3

PR #1224, opened 2020-12-11
Accept const components prop for Trans.

pcorpet:
Typescript bug fix: the `components` prop of `Trans` does not need to be mutable as it won't be modified. I'm changing the types to reflect that. Note that it was already done previously for Typescript <4.1 for arrays.

#### Checklist

- [X] tests are included


jamuhl:
@pedrodurek can you please give a review on this...thank you

pedrodurek (approved):
I agree with the changes made

jamuhl:
was published in react-i18next@11.8.4

Issue #206, opened 2016-10-02
changeLanguage  cause warning


zaaack:
changeLanguage cause warning and has no effect 

warning.js:36
Warning: setState(...): Cannot update during an existing state transition (such as within `render` or another component's constructor). Render methods should be a pure function of props and state; constructor side-effects are an anti-pattern, but can be moved to `componentWillMount`.e @ warning.js:36
r @ warning.js:60
i @ ReactUpdateQueue.js:55
enqueueSetState @ ReactUpdateQueue.js:201
o.setState @ ReactComponent.js:64
onI18nChanged @ translate.js:104
(anonymous function) @ EventEmitter.js:51
emit @ EventEmitter.js:50
addResource @ ResourceStore.js:94
saveMissing @ BackendConnector.js:323
translate @ Translator.js:183
t @ i18next.js:321
fixedT @ i18next.js:313
render @ index.jsx:23
_renderValidatedComponentWithoutOwnerOrContext @ ReactCompositeComponent.js:808
_renderValidatedComponent @ ReactCompositeComponent.js:835
_updateRenderedComponent @ ReactCompositeComponent.js:759
_performComponentUpdate @ ReactCompositeComponent.js:739
updateComponent @ ReactCompositeComponent.js:658
receiveComponent @ ReactCompositeComponent.js:552
receiveComponent @ ReactReconciler.js:126
_updateRenderedComponent @ ReactCompositeComponent.js:761
_performComponentUpdate @ ReactCompositeComponent.js:739
updateComponent @ ReactCompositeComponent.js:658
performUpdateIfNecessary @ ReactCompositeComponent.js:566
performUpdateIfNecessary @ ReactReconciler.js:158
u @ ReactUpdates.js:151
perform @ Transaction.js:138
perform @ Transaction.js:138
perform @ ReactUpdates.js:90
w @ ReactUpdates.js:173
closeAll @ Transaction.js:204
perform @ Transaction.js:151
batchedUpdates @ ReactDefaultBatchingStrategy.js:63
s @ ReactUpdates.js:201
o @ ReactUpdateQueue.js:25
enqueueSetState @ ReactUpdateQueue.js:210
o.setState @ ReactComponent.js:64
onI18nChanged @ translate.js:104
(anonymous function) @ EventEmitter.js:51
emit @ EventEmitter.js:50
addResourceBundle @ ResourceStore.js:125
loaded @ BackendConnector.js:133
(anonymous function) @ BackendConnector.js:240
[...]

PR #1235, opened 2021-01-11
<Trans /> component breaks input values

bripkens:
# Why

Using `<Trans />` to render values that resemble HTML elements breaks
the values.

# What

This PR only adds a test case which reproduces the failure. It does
not fix the issue.

I took a quick look what the root cause is and whether there are
greater security implications (XSS possibilities). The cause seems
to be pretty straightforward. First, `Trans` executes the translation
via `t` and after the successful translation the `string` is
interpreted as HTML.

Screenshot 2021-01-11 at 17 35 14

To me it seems like the strategy should be the other way around –
especially to avoid security issues. Meaning: Parse the translation
string and only afterwards place the values.

Regarding potential security implications: I haven't fully analyzed this.
However, considering how complicated proper sanitization is (the
[DOMPurify](https://github.com/cure53/DOMPurify) project is a good indicator
just how hard this is), I believe this is something that should be verified as well.

### DIFF
--- a/test/trans.render.spec.js
+++ b/test/trans.render.spec.js
@@ -475,3 +475,20 @@ describe('trans should work with misleading overloaded empty elements in compone
     `);
   });
 });
+
+describe('trans should not interpret HTML-like structures in values', () => {
+  const TestComponent = () => (
+    <Trans
+      i18nKey="someKey"
+      defaults="A <strong>{{htmlElement}}</strong> B"
+      values={{ htmlElement: '<head>' }}
+    />
+  );
+
+  it('should render translated string', () => {
+    const { container } = render(<TestComponent />);
+    expect(container.innerHTML).toMatchInlineSnapshot(
+      `"<div>A <strong>&lt;head&gt;</strong> B</div>"`,
+    );
+  });
+});


jamuhl:
passing user input in values you can set:

```js
  <Trans
      i18nKey="someKey"
      defaults="Add the script to the <code>{{htmlElement}}</code>"
      values={{ htmlElement: '<head>' }}
      components={{ code: <code /> }}
tOptions={{interpolation: {
      escapeValue: true
    }}}
    />
```

Not tested but this should escape the input...

bripkens:
@jamuhl: Unfortunately this will cause double escaping as visible by this test case:

```js
describe('trans should not interpret HTML-like structures in values', () => {
  const TestComponent = () => (
    <Trans
      i18nKey="someKey"
      defaults="A {{htmlElement}} B"
      values={{ htmlElement: '<head>' }}
      tOptions={{
        interpolation: {
          escapeValue: true,
        },
      }}
    />
  );

    expect(container.innerHTML).toMatchInlineSnapshot(`"<div>A &amp;lt;head&amp;gt; B</div>"`);
  });
});
```

Note: Matching `container.firstChild` directly as a snapshot will generate a JSX structure doesn't correctly reflect reality. Hence I used `innerHTML` (specifically escaping is funky in snapshots).

This is because the `t` call will already translate `<` to `&lt;`. `&lt;` is then used as a `string` content for a new React element. React then turns `&lt;` into `&amp;lt;`. Overall, I don't see a way to correctly use the `Trans` component with its currently implementation.

jamuhl:
Ah...I didn't check the real output and guessed from the test it does not escape the value.

But:
```js
  <Trans
      i18nKey="someKey"
      defaults="A {{htmlElement}} B"
      values={{ htmlElement: '<head>' }}
    />
```

gives you:

`A <head> B` which is what is expected

----

```js
        <Trans
          i18nKey="someKey"
          defaults="Add the script to the <code>{{htmlElement}}</code>"
          values={{ htmlElement: '<head>' }}
          components={{ code: <code /> }}
        />
```

gives:

`Add the script to the <head></head>` what is not perfect but acceptable and also is no risk as it is escaped

bripkens:
Unfortunately this is only true for a subset of translation strings. Take the following as an example:

```js
describe('trans should not interpret HTML-like structures in values', () => {
  const TestComponent = () => (
    <Trans
      i18nKey="someKey"
      defaults="A <strong>{{htmlElement}}</strong> B"
      values={{ htmlElement: '<head>' }}
    />
  );

    expect(container.innerHTML).toMatchInlineSnapshot(
      `"<div>A <strong>&lt;head&gt;&lt;/head&gt; B</strong></div>"`,
    );
  });
});
```

This is not really the expected output (note the position of the closing `</strong>` which gets forced after `B`).

jamuhl:
That's cause of the used html parser having issues with `<head>` which is non valid HTML.

Guess all this is a compromise -> you can't get all cases working (or not without putting a lot of effort into it)...best might be adding a note -> "don't use HTML like constructs inside value"

jamuhl:
v3.0.0 **BREAKING** changes regex from `/\((\S*)\).*{((.|\n)*)}/` to `/\((\S*)\).*?\[((.|\n)*)\]/` -> from (condition){value} to (condition)[value] to fix [#206](https://github.com/i18next/i18next-intervalPlural-postProcessor/issues/206) and [#221](https://github.com/i18next/i18next-intervalPlural-postProcessor/issues/221)

bripkens:
> Guess all this is a compromise -> you can't get all cases working (or not without putting a lot of effort into it)...best might be adding a note -> "don't use HTML like constructs inside value"

I am just wondering what this means for end-user provided values. End-users could use anything in here. Does this mean the guideline should be: *Never use end-user provided data in values*?

jamuhl:
Only option I currently see to handle this is:

replace`<, >` in values with something like `###opencaret###` before running through HTML parser and replace those again when filling child nodes strings....

bripkens:
@jamuhl: What if `Trans`'s strategy is executed in the exact different order? So:

 1. Get the translation string (so the value behind `i18nKey` / `defaults`)
 2. Parse this string as HTML
 3. Run the `textContent` of each HTML node through `i18n.t`.
 4. Turn this HTML structure into JSX

jamuhl:
you might try...but will be a lot harder and 99% break other cases other users a keen on

jamuhl:
Just to make it clear - there is no way (i know of) to inject some script (no XSS): https://codesandbox.io/s/react-i18next-forked-wnh98?file=/src/app.js

PR #1255, opened 2021-02-14
warn for old wait usage

adrai:
#### Checklist

- [ ] tests are included

### DIFF
--- a/src/useTranslation.js
+++ b/src/useTranslation.js
@@ -18,6 +18,9 @@ export function useTranslation(ns, props = {}) {
     return retNotReady;
   }
 
+  if (i18n.options.react && i18n.options.react.wait !== undefined)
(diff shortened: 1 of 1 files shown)


felixmosh:
Currently suspense is not supported in SSR (tried in next.js), so why `useSuspense` become the standard?

jamuhl:
@felixmosh that change was already made with v10 and looking at next-i18next there is no issue with that as suspense gets only triggered if not fulfilling prerequestitions or alternatively can be turned off

felixmosh:
I'm using next-i18next v4 (the one that requires custom server) with Next.js v10, when disabling `wait` & enabling `useSuspense`, it throws that suspense is not supported in SSR :|

jamuhl:
might be better to ask questions related to next.js in the next-i18next repo...can't really help with this

Issue #1272, opened 2021-03-04
Inform useTranslation hook of replaced i18n instance


medihack:
## 🐛 Bug Report

I use the `I18nextProvider` to distribute the i18n instance in a Blitz.js app with SSR. When switching the language I simply replace the i18n instance in the provider (this is also how [next-i18next](https://github.com/isaachinman/next-i18next/) works). This works when recreating every component, but breaks when just rerendering every component. It is because `t` of `useTranslation` is [cached](https://github.com/i18next/react-i18next/blob/master/src/useTranslation.js#L45).
This also [breaks next-i18next](https://github.com/isaachinman/next-i18next/issues/1023) when using shallow routing. And it is not only a problem with shallow routing, but also with [persistent layouts](https://adamwathan.me/2019/10/17/persistent-layout-patterns-in-nextjs/) as those components in the layout (like app bars) are also not  translated correctly after a page change (and replacing the i18n instance).

## To Reproduce

Replace the i18n instance of a `I18nextProvider` and just force a rerender of all components that use the `useTranslation` hook. I also demonstrated this with a little code change in the example app of next-i18next (https://github.com/medihack/next-i18next).

## Expected behavior

The `useTranslation` hook should somehow check if its provided i18n instance was replaced and not just provide a cached `t` function.

## Your Environment

- *runtime version*: node v15.5
- *i18next version*: 11.8.8
- *os*: Windows WSL

jamuhl:
Not sure who came up with that idea of replacing the instance in the i18nProvider -> correct would be a call to i18n.changeLanguage which would replace the `t` function of useTranslation correctly. 

Changing the instance looks like the badest idea regarding changing the language -> the instance is the base of everything -> all the events are bound to it: https://github.com/i18next/react-i18next/blob/master/src/useTranslation.js#L65

So, replacing the instance and just shallow rerender breaks basically everything...

adrai:
> Not sure who came up with that idea of replacing the instance in the i18nProvider -> correct would be a call to i18n.changeLanguage which would replace the `t` function of useTranslation correctly.
> 
> Changing the instance looks like the badest idea regarding changing the language -> the instance is the base of everything -> all the events are bound to it: https://github.com/i18next/react-i18next/blob/master/src/useTranslation.js#L65
> 
> So, replacing the instance and just shallow rerender breaks basically everything...

@isaachinman this may interest you

isaachinman:
[...]

PR #1273, opened 2021-03-06
Fix allow to replace i18n in provider with useTranslation hook (#1272)

medihack:
-- Prettier seems to be run on committed files as part of the pre-commit husky hook and did some little code changes in useTranslation.js, not my fault ;-)
- [x] tests are included
-- No need for documentation update

### DIFF
--- a/src/useTranslation.js
+++ b/src/useTranslation.js
@@ -76,7 +76,7 @@ export function useTranslation(ns, props = {}) {
     };
   }, [namespaces.join()]); // re-run effect whenever list of namespaces changes
 
-  // t is correctly initialized by useState hook. We only neet to update it after i18n
+  // t is correctly initialized by useState hook. We only need to update it after i18n
   // instance was replaced (for example in the provider).
   const isInitial = useRef(true);
   useEffect(() => {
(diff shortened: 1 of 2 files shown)


src/useTranslation.js
```diff
@@ -68,12 +70,20 @@ export function useTranslation(ns, props = {}) {
...
       isMounted.current = false;
-      if (bindI18n && i18n) bindI18n.split(' ').forEach(e => i18n.off(e, boundReset));
+      if (bindI18n && i18n) bindI18n.split(' ').forEach((e) => i18n.off(e, boundReset));
       if (bindI18nStore && i18n)
-        bindI18nStore.split(' ').forEach(e => i18n.store.off(e, boundReset));
+        bindI18nStore.split(' ').forEach((e) => i18n.store.off(e, boundReset));
     };
   }, [namespaces.join()]); // re-run effect whenever list of namespaces changes
 
+  const isInitial = useRef(true);
```
adrai, line 79: Can you maybe add an inline comment to explain why this is needed?
For future generations ;-)

medihack: Done :-)

adrai: neet -> need

adrai:
lgtm

src/useTranslation.js
medihack: Uups. Fixed.

jamuhl:
thanks to all participating in this fix

was published in react-i18next@11.8.9

Issue #1264, opened 2021-02-25
VS Code Auto-Import Command attempts to import /* of package.


ThomasAunvik:
## 🐛 Bug Report

Visual Studio Code attempts to import `react-i18next` as `react-i18next/*` when attempting to use a function. 
As example here:
![image](https://user-images.githubusercontent.com/9968151/109151591-a85f7900-776a-11eb-950a-d3891805ab63.png)
which does not work and causes this error in the end. 
![image](https://user-images.githubusercontent.com/9968151/109151691-c6c57480-776a-11eb-80b4-e23c3468617c.png)

Removing the extra `/*` works.

## To Reproduce

1. Install VSCode.
2. Open a project in VSCode.
3. `npm i react-i18next`
4. Try to use a function for react-i18next.
5. Use VSCode import / `CTRL+.` keystroak
6. See that vscode is attemtping to import `react-i18next/*`

```js
import { useTranslation } from "react-i18next/*";

const { t } = useTranslation();
```

## Expected behavior

To import the correct path.

```js
import { useTranslation } from "react-i18next";

const { t } = useTranslation();
```

## Attempted "Fixes"
1.
https://github.com/i18next/react-i18next/blob/086246ee653b1ff7679a12dbb6b840c23f22fa0b/package.json#L7-L13
To change to:
```json
"typesVersions": {
  ">=4.1": {
    "*": [
      "src/ts4.1/*"
    ]
  }
},
```

Would fix vscode to import the correct path, but then react refuses to compile cause it can't find module.

2.
Drag items out of the ./src directory in node_modules/react-i18next/ to it's package root.
Edit package json from:

https://github.com/i18next/react-i18next/blob/086246ee653b1ff7679a12dbb6b840c23f22fa0b/package.json#L6-L13
to
```json
"types": "./index.d.ts",
"typesVersions": {
  ">=4.1": {
    "*": [
      "ts4.1/*"
    ]
  }
},
```

This is what seemed to have 'fixed' it up for me.
 
## Your Environment

- *runtime version*: v15.8.0
- *i18next version*: 19.8.4
- *react-i18next version*: 11.8.8
- *os*: Linux

mr00z:
I have the same issue.

pedrodurek:
Thanks @ThomasAunvik, I'll be investigating it

pedrodurek:
[...]

PR #1276, opened 2021-03-12
Move type definition files

pedrodurek:
As typescript team may some time to investigate the issue https://github.com/microsoft/TypeScript/issues/43133. I'm moving the type definitions to address the issue reported here: https://github.com/i18next/react-i18next/issues/1264. 

#### Checklist

- [ ] tests are included


jamuhl:
published in  react-i18next@11.8.10

PR #1283, opened 2021-03-19
Replace html-parse-stringify2 with html-parse-stringify

### LINKED ISSUE #1275: Is react-i18next vulnerable to CVE-2021-23346 ?
## 🐛 Bug Report

Raising as an issue here as I'm not sure this makes sense to ask on SO

- This module uses `html-parse-stringify2@^2.0.1`.  
- [CVE-2021-23346](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-23346) (more info in https://snyk.io/vuln/SNYK-JS-HTMLPARSESTRINGIFY2-1079307) has been raised against that module.  It's not clear whether `html-parse-stringify2` is still maintained and will ever be fixed (ongoing conversation in https://github.com/rayd/html-parse-stringify2/issues/26).
- I _think_ https://github.com/i18next/react-i18next/blob/9f80ddc4f2e4ad13f57e12ad4ad752a35c7ee496/src/Trans.js#L136 is the only place that `html-parse-stringify2` is used, where the string to be parsed is explicitly wrapped in `<0>${interpolatedString}</0>`
- I have not yet managed to find a value for `${interpolatedString}` that triggers the ReDoS bug in `html-parse-stringify2`, and looking at the bug in the RegEx in `html-parse-stringify2` I'm pretty sure there isn't one, but would like an official view from the maintainers on whether CVE-2021-23346 is definitely not applicable.

## To Reproduce

If you want to see the bug in `html-parse-stringify2` then run...

```js
const p = require('html-parse-stringify2')

console.log('parsing')
p.parse("<!'''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''!")
console.log('parsed')
```

This does not complete in a timely manner.

Running the following does complete quickly, but maybe I just can't think of a "bad" input yet.

```js
const p = require('html-parse-stringify2')

console.log('parsing')
p.parse("<0><!'''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''!></0>")
console.log('parsed')
```

## Expected behavior

Hopefully you confirm this is not applicable

## Your Environment

- *runtime version*: Node v14
- *i18next version*: i.e. 14.1.1
- *os*: Linux

kachkaev:
Closes #1275

#### Checklist

- ~tests are included~
- ~documentation is changed or added~

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5549,18 +5549,11 @@
       "dev": true
     },
     "html-parse-stringify": {
-      "version": "2.0.2",
-      "resolved": "https://registry.npmjs.org/html-parse-stringify/-/html-parse-stringify-2.0.2.tgz",
-      "integrity": "sha512-jcyzMGN3C/aO8fLf0Yjz6uBtP0diUNDG9g0rgVdE51mo8b2O3UOlar8Epc/Z7gE6jPGAvU41v9uE+g+R8izhtA==",
+      "version": "2.0.3",
+      "resolved": "https://registry.npmjs.org/html-parse-stringify/-/html-parse-stringify-2.0.3.tgz",
+      "integrity": "sha512-a0Uj07o4HaLPZ46KaEyR3VDwBYIUEdH9dv/iHs2T5ZsD1zwdvfGBg2zhQclGtCF9MoZ5fUi3RIkWqM5VRO/uOw==",
       "requires": {
         "void-elements": "3.1.0"
-      },
-      "dependencies": {
-        "void-elements": {
-          "version": "3.1.0",
-          "resolved": "https://registry.npmjs.org/void-elements/-/void-elements-3.1.0.tgz",
-          "integrity": "sha1-YU9/v42AHwu18GYfWy9XhXUOTwk="
-        }
       }
     },
     "http-signature": {
@@ -11049,6 +11042,11 @@
         "core-util-is": "1.0.2",
         "extsprintf": "^1.2.0"
       }
+    },
+    "void-elements": {
+      "version": "3.1.0",
+      "resolved": "https://registry.npmjs.org/void-elements/-/void-elements-3.1.0.tgz",
+      "integrity": "sha1-YU9/v42AHwu18GYfWy9XhXUOTwk="
     },
     "w3c-hr-time": {
       "version": "1.0.2",
--- a/package-lock.json
+++ b/package-lock.json
@@ -5548,12 +5548,19 @@
       "integrity": "sha512-H2iMtd0I4Mt5eYiapRdIDjp+XzelXQ0tFE4JS7YFwFevXXMmOp9myNrUvCg0D6ws8iqkRPBfKHgbwig1SmlLfg==",
       "dev": true
     },
-    "html-parse-stringify2": {
-      "version": "2.0.1",
-      "resolved": "https://registry.npmjs.org/html-parse-stringify2/-/html-parse-stringify2-2.0.1.tgz",
-      "integrity": "sha1-3FZwtyksoVi3vJFsmmc1rIhyg0o=",
+    "html-parse-stringify": {
+      "version": "2.0.2",
+      "resolved": "https://registry.npmjs.org/html-parse-stringify/-/html-parse-stringify-2.0.2.tgz",
+      "integrity": "sha512-jcyzMGN3C/aO8fLf0Yjz6uBtP0diUNDG9g0rgVdE51mo8b2O3UOlar8Epc/Z7gE6jPGAvU41v9uE+g+R8izhtA==",
       "requires": {
-        "void-elements": "^2.0.1"
+        "void-elements": "3.1.0"
+      },
+      "dependencies": {
+        "void-elements": {
+          "version": "3.1.0",
+          "resolved": "https://registry.npmjs.org/void-elements/-/void-elements-3.1.0.tgz",
+          "integrity": "sha1-YU9/v42AHwu18GYfWy9XhXUOTwk="
+        }
       }
     },
     "http-signature": {
@@ -11042,11 +11049,6 @@
         "core-util-is": "1.0.2",
         "extsprintf": "^1.2.0"
       }
-    },
-    "void-elements": {
-      "version": "2.0.1",
-      "resolved": "https://registry.npmjs.org/void-elements/-/void-elements-2.0.1.tgz",
-      "integrity": "sha1-wGavtYK7HLQSjWDqkjkulNXp2+w="
     },
     "w3c-hr-time": {
       "version": "1.0.2",
--- a/package-lock.json
+++ b/package-lock.json
@@ -5549,9 +5549,9 @@
       "dev": true
     },
     "html-parse-stringify": {
-      "version": "2.1.1",
-      "resolved": "https://registry.npmjs.org/html-parse-stringify/-/html-parse-stringify-2.1.1.tgz",
-      "integrity": "sha512-w+9Q9DdT1RoiM95SscK8c04IxHaf07RbORlMNvaXXOtLrCm3tm73P/cCbJoWNKRbLvmQfEDNjRhAu8yfWYsxxg==",
+      "version": "3.0.0",
+      "resolved": "https://registry.npmjs.org/html-parse-stringify/-/html-parse-stringify-3.0.0.tgz",
+      "integrity": "sha512-TrTKp/U0tACrpqalte/VhxepqMLii2mOfC8iuOt4+VA7Zdi6BUKKqNJvEsO17Cr3T3E7PpqLe3NdLII6bcYJgg==",
       "requires": {
         "void-elements": "3.1.0"
       }
--- a/package-lock.json
+++ b/package-lock.json
@@ -5549,9 +5549,9 @@
       "dev": true
     },
     "html-parse-stringify": {
-      "version": "2.1.0",
-      "resolved": "https://registry.npmjs.org/html-parse-stringify/-/html-parse-stringify-2.1.0.tgz",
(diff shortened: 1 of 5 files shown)


kachkaev:
The upstream issue is now fixed, but a bunch of tests for `<Trans />` fail. Example:

```
FAIL  test/trans.render.list.spec.js (28.115s)
  ● Trans should render nested components › should render dynamic ul as components property

    expect(received).toMatchInlineSnapshot(snapshot)

    Snapshot name: `Trans should render nested components should render dynamic ul as components property 1`

    - Snapshot
    + Received

    - <div>
    -   Result should be a list: 
    -   <ul>
    -     <li>
    -       li1
    -     </li>
    -     <li>
    -       li2
    -     </li>
    -   </ul>
    - </div>
    + <div />
```

Any thoughts?

adrai:
Practically all Trans tests are failing...
I suspect the generated ast looks completely different.... https://github.com/i18next/react-i18next/blob/master/src/Trans.js#L136

adrai:
https://github.com/rayd/html-parse-stringify2/pull/27

kachkaev:
@adrai thanks for your alternative solution! To increase our chances for a fast resolution, do you know what I could change in this PR for the tests to pass? I haven’t used `html-parse-stringify` or `html-parse-stringify2` before, so am not familiar with their differences or expectations. Intuitively, it feels better to return back to `html-parse-stringify` from the fork.

jamuhl:
@kachkaev guess you have to write new snapshots -> and manually check first if received output still matches expected and the changes are pure formatting and not wrong html

kachkaev:
Hmmm here’s the catch. When we call

```
  const ast = HTML.parse(`<0>${interpolatedString}</0>`);
```

using `html-parse-stringify@2.0.3`, the result is an empty array. This is [because of](https://github.com/HenrikJoreteg/html-parse-stringify/blob/master/src/parse.js#L1):

```ts
const tagRE = /<[a-zA-Z\-\!\/](?:"[^"]*"|'[^']*'|[^'">])*>/g
```

Thus, tags like `<0>`, `<1>` etc. are ignored.

If https://github.com/rayd/html-parse-stringify2/pull/27 is merged, we’ll end up with the same problem.

adrai:
PR https://github.com/rayd/html-parse-stringify2/pull/27 is adapted

kachkaev:
Two blockers on the path to `html-parse-stringify`:

- https://github.com/HenrikJoreteg/html-parse-stringify/pull/43
- https://github.com/HenrikJoreteg/html-parse-stringify/pull/44

kachkaev:
bump

kachkaev:
@HenrikJoreteg has just released [`html-parse-stringify`](https://github.com/HenrikJoreteg/html-parse-stringify)`@2.1.0` – it includes https://github.com/HenrikJoreteg/html-parse-stringify/pull/43 so we are half-way there!

Once https://github.com/HenrikJoreteg/html-parse-stringify/pull/41 is merged and released, all tests should be passing and we’ll be able to switch from `html-parse-stringify2` to `html-parse-stringify` 👍

kachkaev:
I went into my local `nodes_modules` and manually tweaked `html-parse-stringify/dist` code so that it included https://github.com/HenrikJoreteg/html-parse-stringify/pull/41. Interestingly, I noticed that some snapshots ended up failing, so I updated them. Is this a regression or just an artifact in snapshots? 🤔 See diff.

adrai:
> Is this a regression or just an artifact in snapshots? 🤔 If it is a regression, is it something to fix here or upstream? Potentially, we might have a lost space between `</closingTag> <openingTag>`.

No errors on master:
image

kachkaev:
No errors on master is expected, because in there we’ve got `"html-parse-stringify2": "^2.0.1"`. This PR replaces that dependency with `html-parse-stringify`, which either has a slightly different shape of an AST or misses a space in an AST in the first place. I don’t have a solution on top of my head right now, so if anyone could do some extra digging, that’d be great!

Where I’m going is that we might need to either submit an extra patch to `html-parse-stringify` in addition to https://github.com/HenrikJoreteg/html-parse-stringify/pull/41 or maybe amend AST parsing here.

https://github.com/rayd/html-parse-stringify2/pull/27 is still a good alternative solution and I’m happy with any that can land first. The main goal here is to close #1275 🙂

kachkaev:
Thanks to @HenrikJoreteg’s work on merging https://github.com/HenrikJoreteg/html-parse-stringify/pull/41 and releasing of `2.1.1`, we have CI passing now! However, for this to work I had to remove extra new lines from the snapshots, which we’ll need to investigate. My gut feeling is that there’s another small bug in `html-parse-stringify`, in particular to do with treating `</a> <b>` as `</a><b>`. The alternative cause could be to do with some AST handling here (which would be easier to fix).

To confirm or deny that we’ll need to add some extra test cases upstream. I’m not sure when exactly I’ll be able to do this, so if anyone wants to help, feel free to 🙌

HenrikJoreteg:
If you could give me a sample of the failing input I might be able to look
into it.

adrai:
I had a quick look at the generated ast of this failing test: https://github.com/i18next/react-i18next/blob/master/test/trans.render.spec.js#L129

Just a simple `console.log(JSON.stringify(ast, null, 2))` here: https://github.com/i18next/react-i18next/blob/master/src/Trans.js#L137

The output of html-parse-stringify differs from html-parse-stringify2 in the following way:

This text element is missing in html-parse-stringify before the br element:
```json
      {
        "type": "text",
        "content": " "
      }
```


**html-parse-stringify2:**
```json
[
    {
        "type": "tag",
        "name": "0",
        "voidElement": false,
        "attrs": {},
        "children": [
            {
                "type": "tag",
                "name": "strong",
                "voidElement": false,
                "attrs": {},
                "children": [
                    {
                        "type": "text",
                        "content": "Go"
                    }
                ]
            },
            {
                "type": "text",
                "content": " "
            },
            {
                "type": "tag",
                "name": "br",
                "voidElement": true,
                "attrs": {},
                "children": []
            },
            {
                "type": "tag",
                "name": "1",
                "voidElement": false,
                "attrs": {},
                "children": [
                    {
                        "type": "text",
                        "content": "there"
                    }
                ]
            },
            {
                "type": "text",
                "content": "."
            }
        ]
    }
]
```


**html-parse-stringify:**
```json
[
    {
        "type": "tag",
        "name": "0",
        "voidElement": false,
        "attrs": {},
        "children": [
            {
                "type": "tag",
                "name": "strong",
                "voidElement": false,
                "attrs": {},
                "children": [
                    {
                        "type": "text",
                        "content": "Go"
                    }
                ]
            },
            {
                "type": "tag",
                "name": "br",
                "voidElement": true,
                "attrs": {},
                "children": []
            },
            {
                "type": "tag",
                "name": "1",
                "voidElement": false,
                "attrs": {},
                "children": [
                    {
                        "type": "text",
                        "content": "there"
                    }
                ]
            },
            {
                "type": "text",
                "content": "."
            }
        ]
    }
]
```

pavanjava:
Hi Team,

react-i18next transitively depends on html-parse-stringify2 and has got the vulnerability #CVE-2021-23346. when can we expect the new release of react-i18next with the corrections ? is this planned for near future. curious to ask after seeing all the work done on going back to fork.

kachkaev:
@pavanjava, the goal of this PR is to mitigate CVE-2021-23346 – you are right about it. The ETA for a new `react-i18next` release depends on how fast we will resolve this bug in `html-parse-stringify`: https://github.com/i18next/react-i18next/pull/1283#issuecomment-812346154. Feel free to help!

adrai:
Tried to create a PR: https://github.com/HenrikJoreteg/html-parse-stringify/pull/45
Really not sure if this is the correct way to do so... 
this in combination with passing `{ respectWhitespace: true }` in react-i18next:
image
could fix the tests...
But honestly I do not even know if having or not having that whitespace matters?

kachkaev:
> Tried to create a PR

Great stuff, thanks @adrai!

>  I do not even know if having or not having that whitespace matters?

I _think_ it might matter. If we have `<b>hello</b> <i>world</i>` and the whitespace token is removed, we end up with `helloworld`.

alon24:
any news?

HenrikJoreteg:
I think I'm the hangup, i'm trying to sort out if it should be implemented as an option (as was done) in the PR or if it should just be the normal behavior to always preserve it.  I'd rather not make it an option unless we can think of a reason why anyone would want to not have the whitespace.

kachkaev:
I second an option that avoids having options. If anyone wants to not have space between two tags, they can remove it in the source string 🙂

adrai:
> I second an option that avoids having options. If anyone wants to not have space between two tags, they can remove it in the source string 🙂

new major version of html-parse-stringify ;-) https://github.com/HenrikJoreteg/html-parse-stringify/pull/45#issuecomment-817323584

adrai:
Good news: https://github.com/HenrikJoreteg/html-parse-stringify/pull/46#issuecomment-817816776

adrai:
@kachkaev you can choose:
1. Do you want to update this PR?
2. I'll do it directly on master and create a new version?

kachkaev:
Great news! 🎉 Happy to update the PR shortly 🙌

kachkaev:
:shipit:

adrai:
patch?

adrai:
🎉🎉🎉🎉🎉🎉 PUBLISHED v11.8.13 🎉🎉🎉🎉🎉🎉🎉

PR #1286, opened 2021-03-25
refactor: remove unneeded object

TQjona:
Minor optimizations

Works by switching to `useState(() => initialValue)` and `setState((prevState) => nextState)`

#### Checklist

- [x] tests are included
- [x] documentation is changed or added

### DIFF
--- a/src/useTranslation.js
+++ b/src/useTranslation.js
@@ -40,11 +40,9 @@ export function useTranslation(ns, props = {}) {
 
   // binding t function to namespace (acts also as rerender trigger)
   function getT() {
-    return {
-      t: i18n.getFixedT(null, i18nOptions.nsMode === 'fallback' ? namespaces : namespaces[0]),
-    };
+    return i18n.getFixedT(null, i18nOptions.nsMode === 'fallback' ? namespaces : namespaces[0]);
   }
-  const [t, setT] = useState(getT()); // seems we can't have functions as value -> wrap it in obj
+  const [t, setT] = useState(getT);
 
   const isMounted = useRef(true);
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


jamuhl:
Will now unneeded -> does it make a difference?

TQjona:
This saves a few bytes (~16) in bundle size and some heap memory (approx. 64 bytes per element using `useTranslation`) due to the object being removed. Theoretically you should also get better performance due to less references on the heap (less work for the garbage collector) and less operations in the JS code, but in the real world those improvements are too small to measure. React can't even handle the amount of elements required to make this effect visible.

I'd still consider merging the PR due to the widespread use of this library, although I wouldn't create an extra release with just this small change.

src/useTranslation.js
```diff
@@ -40,11 +40,9 @@ export function useTranslation(ns, props = {}) {
 
   // binding t function to namespace (acts also as rerender trigger)
   function getT() {
-    return {
-      t: i18n.getFixedT(null, i18nOptions.nsMode === 'fallback' ? namespaces : namespaces[0]),
-    };
+    return i18n.getFixedT(null, i18nOptions.nsMode === 'fallback' ? namespaces : namespaces[0]);
   }
-  const [t, setT] = useState(getT()); // seems we can't have functions as value -> wrap it in obj
```
adrai, line 47: what was the reason for this comment?

TQjona: I assume the original author of the code just tried `useState(getT())` with `getT()` returning the i18n.t function. React detected that the parameter passed to useState was a function and thought it was a lazy-init function (`useState(() => initialValue)`). So it was calling i18n.t with no args and used the return value as initial value, which was probably null or undefined.

Basically to store a function in a state you have to have a function which returns the function, e.g. `useState(() => (args) => console.log(args))` and `setState(() => (args) => console.log(args))`

adrai:
lgtm

jamuhl:
thank you...was published in react-i18next@11.8.12

PR #1300, opened 2021-04-15
Dependabot configuration

mmorel-35:
Define dependabot to keep npm dependencies and github action up-to-date

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+version: 2
+updates:
+- package-ecosystem: npm
+  directory: /
+  target-branch: master
+  schedule:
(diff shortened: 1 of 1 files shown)


adrai:
personally not a fan of automatic dependency updates

mmorel-35:
Hi @adrai , why do you think that? With dependabot, the update is proposed automatically, it creates a PR and passes your ci testing like any other PR you woul receive . You can always reject the PR if you don't want it.

adrai:
Just my personal opinion...
usually when creating a new version I check for new dependency versions with npm-check-update...

mmorel-35:
I see...

Issue #1308, opened 2021-04-27
Introduce a new prop in the Trans component called keyComponents.


iamrane:
## 🚀 Feature Proposal
I would like to introduce a new prop to the Trans component called keyComponents. And what it does is that it wrap a key in a translation string with an element.

## Motivation
1. Unnecessary tags in the translations
Alot of times we end up with `Hello <firstname>{firstname}</firstname>` types of translations. A wrapper around a key with the exact same name. These feels unnecessary in these kind of places.

2. More reusable keys
I can use the same i18nkey on several places. Sometimes with a styling and sometimes without. (see example)

3. The styling ends up in the front end code, and not in the data.
Often the people who do translation they only care for the translations. They don't always know where it should be used or how it should look. With components styling ends up in the translations and I think it's not optimal. With keyComponents the styling will end up in the front end layer where it belongs.

## Example

So instead of writing
```js
<Trans
    t={t}
    i18nKey="myi18nkey"
    defaults="Hello <firstname>{firstname}</firstname>"
    values={{ firstname: 'Christian' }}
    components={{ firstname: <span style={{ color: 'red' }} /> }}
/>
```

I imagine something like this:
```js
<Trans
    t={t}
    i18nKey="myi18nkey"
    defaults="Hello {firstname}"
    values={{ firstname: 'Christian' }}
    keyComponents={{ firstname: <span style={{ color: 'red' }} /> }}
/>
```

Then I can use the same i18nKey in other places where I don't want the highlight styling
```js
{t('myi18nKey', { defaultValue: 'Hello {firstname}', firstname: 'Christian' })}
```

Im not saying we should replace components, because there are places where that makes sense. But for me keyComponents would replace components 9 times of 10.

adrai:
Would you like to try a Pull Request to address this? Remember to add unit tests.

iamrane:
I will have a try. Need to understand how the Trans component works. But will have a look.

jamuhl:
Why not just:

`<span style={{ color: 'red' }}>{t('myi18nkey', { firstname: 'Christian' })}</span>`

or if you got to use Trans (not sure why people use Trans for simple strings)

```js
<span style={{ color: 'red' }}>
<Trans
    t={t}
    i18nKey="myi18nkey"
    defaults="Hello {firstname}"
    values={{ firstname: 'Christian' }}
/>
</span>
```

or using parent props

```js
<Trans
    t={t}
parent='span'
style={ color: 'red' }
[...]

PR #1309, opened 2021-04-29
Added support for keyComponents prop in Trans component

iamrane:
#### Checklist

- [x] tests are included


iamrane:
Is this exceeding the limit of lines per file something I need to deal with. Will lead to me refactoring code I didn’t write. Or can you approve a PR without me resolving this issue?

jamuhl:
I don't like the approach of mutating the value in the resourcestore to much...why not just convert that stuff here: https://github.com/i18next/react-i18next/blob/master/src/Trans.js#L114 before the interpolation replaces those interpolationVariables?

calling the new prop: `keyComponents` not sure if that is not misleading (i18n keys, what are key components) -> eventual something like `interpolationWrappers` would be better?

Beside that - not sure what the end of discussion https://github.com/i18next/react-i18next/issues/1308#issuecomment-831017220 is - perhaps a few more community opinions would help...

iamrane:
> I don't like the approach of mutating the value in the resourcestore to much...why not just convert that stuff here: https://github.com/i18next/react-i18next/blob/master/src/Trans.js#L114 before the interpolation replaces those interpolationVariables?
> 
> calling the new prop: `keyComponents` not sure if that is not misleading (i18n keys, what are key components) -> eventual something like `interpolationWrappers` would be better?
> 
> Beside that - not sure what the end of discussion [#1308 (comment)](https://github.com/i18next/react-i18next/issues/1308#issuecomment-831017220) is - perhaps a few more community opinions would help...

Yeah I agree on that changing the resource feels "bad". I didn't deep dive in the code that hard. Thanks for pointing out where in the code I can have a look. Will look into it later on.

iamrane:
So i updated the Trans component now so the I don't manipulate the resource.
And I also moved in the manipulation to where @jamuhl guided me.

iamrane:
Hi!
I will close this PR, i values is not really a simple key-value relation it can be so much more. You can add formatters in the resource string etc. So its not really techical possible to do it how I want it im afraid.

I will need to use components instead. Thanks any way.

PR #1310, opened 2021-05-01
fix #1299: add support for icu macro in typescript

### LINKED ISSUE #1299: typescript types for ICU macros `Plural` `Select` and `SelectOrdinal` are missing
## 🐛 Bug Report

The types are missing, so you can't use these ICU macros in Typescript.

## Expected behavior

These types work on my local for all possible props, but require typescript 4.2+. They will error if you don't pass in `other`, and will allow any `$#` prop (`$0`, `$12345` are both valid) on `Plural`/`SelectOrdinal` but strongly type is as a string or a ReactElement (to allow `<Trans>`).

```js
export interface PluralSubProps<
  K extends TFuncKey<N> extends infer A ? A : never,
  N extends Namespace = DefaultNamespace,
> {
  children?: never;
  i18nKey?: K;
  i18n?: i18n;
  ns?: N;
  count: number;
  zero?: string | React.ReactElement;
  one?: string | React.ReactElement;
  two?: string | React.ReactElement;
  few?: string | React.ReactElement;
  many?: string | React.ReactElement;
  other: string | React.ReactElement;
}

type PluralProps<
  T,
  K extends TFuncKey<N> extends infer A ? A : never,
  N extends Namespace = DefaultNamespace,
> = {
  [P in keyof T]: P extends keyof PluralSubProps<K, N> ? PluralSubProps<K, N>[P] :
  P extends `$${number}` ? string | React.ReactElement : never;
}

interface SelectSubProps {
  [key: string]: string | React.ReactElement;
}

interface SelectRequiredProps<
K extends TFuncKey<N> extends infer A ? A : never,
N extends Namespace = DefaultNamespace,
> {
  children?: never;
  i18nKey?: K;
  i18n?: i18n;
  ns?: N;
  other: string | React.ReactElement;
}

type SelectProps<
K extends TFuncKey<N> extends infer A ? A : never,
N extends Namespace = DefaultNamespace,
> = SelectSubProps & SelectRequiredProps<K, N>;

export function Plural<
  T,
  K extends TFuncKey<N> extends infer A ? A : never,
  N extends Namespace = DefaultNamespace,
>(props: PluralProps<T, K, N>): React.ReactElement;

export function SelectOrdinal<
  T,
  K extends TFuncKey<N> extends infer A ? A : never,
  N extends Namespace = DefaultNamespace,
>(props: PluralProps<T, K, N>): React.ReactElement;

export function Select<
K extends TFuncKey<N> extends infer A ? A : never,
N extends Namespace = DefaultNamespace,
>(props: SelectProps<K, N>): React.ReactElement;
```

cellog:
#### Checklist

- [ ] tests are included

### DIFF
--- a/ts4.1/icu.d.ts
+++ b/ts4.1/icu.d.ts
@@ -1,0 +1,85 @@
+declare module 'react-i18next/icu.macro' {
+  type Namespace = import('.').Namespace;
+  type DefaultNamespace = import('.').DefaultNamespace;
+  type Resources = import('.').Resources;
+  type TFuncKey<N extends Namespace = DefaultNamespace, T = Resources> = import('.').TFuncKey<N, T>;
+  type i18n = import('i18next').i18n;
+
+  export interface PluralSubProps<
+    K extends TFuncKey<N> extends infer A ? A : never,
+    N extends Namespace = DefaultNamespace
+  > {
+    children?: never;
+    i18nKey?: K;
+    i18n?: i18n;
+    ns?: N;
+    count: number;
+    zero?: string | React.ReactElement;
+    one?: string | React.ReactElement;
+    two?: string | React.ReactElement;
+    few?: string | React.ReactElement;
+    many?: string | React.ReactElement;
+    other: string | React.ReactElement;
+  }
+
+  type PluralProps<
+    T,
+    K extends TFuncKey<N> extends infer A ? A : never,
+    N extends Namespace = DefaultNamespace
+  > = {
+    [P in keyof T]: P extends keyof PluralSubProps<K, N>
+      ? // support the standard properties of Plural
+        PluralSubProps<K, N>[P]
+      : // this supports infinite $0={..} or $123={..}
+      // technically it also supports $-1={..} and $2.3={..} but we don't need to
+      // worry since that's invalid syntax.
+      P extends `$${number}`
+      ? string | React.ReactElement
+      : never;
+  };
+
+  interface SelectSubProps {
+    [key: string]: string | React.ReactElement;
+  }
+
+  interface SelectRequiredProps<
+    K extends TFuncKey<N> extends infer A ? A : never,
+    N extends Namespace = DefaultNamespace
+  > {
+    children?: never;
+    i18nKey?: K;
+    i18n?: i18n;
+    ns?: N;
+    other: string | React.ReactElement;
+  }
+
+  // defining it this way ensures that `other` is always defined, but allows
+  // unlimited other select types.
+  type SelectProps<
+    K extends TFuncKey<N> extends infer A ? A : never,
+    N extends Namespace = DefaultNamespace
+  > = SelectSubProps & SelectRequiredProps<K, N>;
+
+  export function Plural<
+    T,
+    K extends TFuncKey<N> extends infer A ? A : never,
+    N extends Namespace = DefaultNamespace
+  >(props: PluralProps<T, K, N>): React.ReactElement;
+
+  export function SelectOrdinal<
+    T,
+    K extends TFuncKey<N> extends infer A ? A : never,
+    N extends Namespace = DefaultNamespace
+  >(props: PluralProps<T, K, N>): React.ReactElement;
+
+  export function Select<
+    K extends TFuncKey<N> extends infer A ? A : never,
+    N extends Namespace = DefaultNamespace
+  >(props: SelectProps<K, N>): React.ReactElement;
+
+  export function Trans<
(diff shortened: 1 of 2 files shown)


ts4.1/icu.d.ts
```diff
@@ -0,0 +1,85 @@
+declare module 'react-i18next/icu.macro' {
+  type Namespace = import('.').Namespace;
+  type DefaultNamespace = import('.').DefaultNamespace;
+  type Resources = import('.').Resources;
+  type TFuncKey<N extends Namespace = DefaultNamespace, T = Resources> = import('.').TFuncKey<N, T>;
+  type i18n = import('i18next').i18n;
```
pedrodurek, line 6: Everything inside the module will be exported (even without the `export` keyword), I don't think we have an `i18n` under `react-i18next`. If you don't want to re-export all these types, you should import them outside the module.
```suggestion
import {Namespace, DefaultNamespace, Resources, TFuncKey} from '.';
import {i18n} from 'i18next';

declare module 'react-i18next/icu.macro' {
```

ts4.1/icu.d.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  export interface PluralSubProps<
+    K extends TFuncKey<N> extends infer A ? A : never,
+    N extends Namespace = DefaultNamespace
+  > {
+    children?: never;
```
pedrodurek, line 12: Here we can simply remove children

ts4.1/icu.d.ts
```diff
@@ -0,0 +1,85 @@
+declare module 'react-i18next/icu.macro' {
+  type Namespace = import('.').Namespace;
+  type DefaultNamespace = import('.').DefaultNamespace;
+  type Resources = import('.').Resources;
+  type TFuncKey<N extends Namespace = DefaultNamespace, T = Resources> = import('.').TFuncKey<N, T>;
+  type i18n = import('i18next').i18n;
+
+  export interface PluralSubProps<
```
pedrodurek, line 8: As I mentioned above all types inside the module will be exported so we don't need the `export` keyword

ts4.1/icu.d.ts
cellog: the intent here was to prevent children. In other words, typescript will error if you pass children in, as a way of signaling "children will be ignored"

ts4.1/icu.d.ts
cellog: when I did it this way, the types stopped working in VS Code, and it also failed to type them at all (they became `any`). Can you confirm?

pedrodurek: yeah, it works perfectly, but you also need to rename the file declaration to `uci.macro.d.ts`, otherwise, it won't find the module.

Screenshot 2021-05-08 at 1 14 01 PM

ts4.1/icu.d.ts
```diff
@@ -0,0 +1,85 @@
...
+    K extends TFuncKey<N> extends infer A ? A : never,
+    N extends Namespace = DefaultNamespace
+  >(props: SelectProps<K, N>): React.ReactElement;
+
+  export function Trans<
+    K extends TFuncKey<N> extends infer A ? A : never,
+    N extends Namespace = DefaultNamespace,
+    E extends Element = HTMLDivElement
+  >(props: import('.').TransProps<K, N, E>): React.ReactElement;
+}
```
pedrodurek, line 85: You can re-export the `Trans` function, but it needs to be done outside the module.
```suggestion
}

export { Trans };
```

ts4.1/icu.d.ts
pedrodurek: I believe this will happen for `Select` component only 🤔 , because you're intersecting its props with `SelectSubProps`

cellog: true, I'll add it to PluralProps to solve that oversight

ts4.1/icu.d.ts
cellog: FYI, tslint is going nuts now. Should we be concerned? I'm going to push this as you recommend, and then we can decide this question.

Screen Shot 2021-05-08 at 9 09 15 PM

pedrodurek: So you can add the export again if you want

pedrodurek (approved):
LGTM!!! 🎉 🙌

cellog:
awesome! I'm going to put up the other PR idea in a draft PR. I will rebase it on master once this is merged, so I can add the typescript types for it too. Nice doing bidness wit you :)

PR #1328, opened 2021-06-04
Allow for Custom DefaultNS typing

### LINKED ISSUE #1302: Declaring a different defaultNS than "translation" allows unexpected cases on t function. [Typescript]
## 🐛 Bug Report

Declaring a different defaultNS than "translation" allows unexpected cases on t function.

Related issues: https://github.com/i18next/react-i18next/issues/1301

I think the issue might be related to [this line](https://github.com/i18next/react-i18next/blob/master/ts4.1/index.d.ts#L161
) since 'translation' is hardcored.
## To Reproduce

[codesandbox](https://codesandbox.io/live/h4thw4o?file=/src/App.tsx);

~~~tsx
// i18nConfig

i18n.use(initReactI18next).init({
  defaultNS: "ns1",
  resources: {
  sk: {
    ns1: {
      a: "translated A"
    },
    ns2: {
      b: "translated B"
    },
    ns3: {
      c: "translated C"
    }
  }
} 
});
~~~

~~~tsx
// App.tsx
  const { t } = useTranslation();
  ...
  <h4>{t("c")}</h4> {/* It should not compile */}

~~~

## Expected behavior

When a specifing a different defaultNS than "translation", TS compilation should not allow cases that are not supported by JS implementation.

## Your Environment

- *i18next version*: i.e. 20.2.1
- *react-i18next version*: 11.8.13

JCQuintas:
## Description

This PR allows the declaration of custom `defaultNS` types and should fix #1302 
It also implements a new way of typing the `Resources` by using the `resources` key inside `CustomTypeOptions` instead of the previous API, `Resources` has been marked as deprecated as well

## Usage

```ts
// i18n.ts
i18n.use(initReactI18next).init({
  defaultNS: 'ns1'
  ...
});

// react-i18next.d.ts
declare module 'react-i18next' {
  interface CustomTypeOptions {
    defaultNS: 'ns1';
    resources: {
      ns1: typeof ns1;
      ns2: typeof ns2;
    };
  }
}
```

#### Checklist

- [x] tests are included

### DIFF
--- a/ts4.1/index.d.ts
+++ b/ts4.1/index.d.ts
@@ -24,15 +24,40 @@ type Subtract<T extends K, K> = Omit<T, keyof K>;
  */
 export interface Resources {}
 /**
- * This interface can be augmented by users in case they define a custom `defaultNS`.
+ * This interface can be augmented by users to add types to `react-i18next`. It accepts a `defaultNS` and `resources` properties.
  *
+ * Usage:
  * ```ts
- * interface DefaultNS { name: 'myCustomNamespace' }
+ * // react-i18next.d.ts
+ * import 'react-i18next';
+ * declare module 'react-i18next' {
+ *   interface CustomTypeOptions {
+ *     defaultNS: 'custom';
+ *     resources: {
+ *       custom: {
+ *         foo: 'foo';
+ *       };
+ *     };
+ *   }
+ * }
  * ```
  */
-export interface DefaultNS {}
+export interface CustomTypeOptions {}
 
-type Fallback<F, T = keyof Resources> = [T] extends [never] ? F : T;
+type MergeBy<T, K> = Omit<T, keyof K> & K;
+
+type TypeOptions = MergeBy<
+  {
+    defaultNS: 'translation';
+    resources: Resources;
+  },
+  CustomTypeOptions
+>;
+
+type DefaultResources = TypeOptions['resources'];
+type DefaultNamespace<T = TypeOptions['defaultNS']> = T extends Fallback<string> ? T : string;
+
+type Fallback<F, T = keyof DefaultResources> = [T] extends [never] ? F : T;
 
 export type Namespace<F = Fallback<string>> = F | F[];
 
@@ -98,13 +123,16 @@ type NormalizeMultiReturn<T, V> = V extends `${infer N}:${infer R}`
     : never
   : never;
 
-export type TFuncKey<N extends Namespace = DefaultNamespace, T = Resources> = N extends (keyof T)[]
+export type TFuncKey<
+  N extends Namespace = DefaultNamespace,
+  T = DefaultResources
+> = N extends (keyof T)[]
   ? NormalizeMulti<T, N[number]>
   : N extends keyof T
   ? Normalize<T[N]>
   : string;
 
-export type TFuncReturn<N, TKeys, TDefaultResult, T = Resources> = N extends (keyof T)[]
+export type TFuncReturn<N, TKeys, TDefaultResult, T = DefaultResources> = N extends (keyof T)[]
   ? NormalizeMultiReturn<T, TKeys>
   : N extends keyof T
   ? NormalizeReturn<T[N], TKeys>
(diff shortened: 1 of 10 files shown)


JCQuintas:
I'm not sure how I can test this, but I'm open to suggestions ofc.

Also, should I just create a new MR on https://github.com/i18next/react-i18next-gitbook for documentation or should I wait?

pedrodurek:
Thanks @JCQuintas, I'll be taking a look and perform some tests this week

ts4.1/index.d.ts
```diff
@@ -158,7 +166,12 @@ type UseTranslationResponse<N extends Namespace> = [TFunction<N>, i18n, boolean]
   ready: boolean;
 };
 
-type DefaultNamespace<T = 'translation'> = T extends Fallback<string> ? T : string;
+type TranslationNS = 'translation';
+type NameKey = 'name';
+
+type ExtractName<T> = T extends { [key in NameKey]: unknown } ? T[NameKey] : TranslationNS;
+
+type DefaultNamespace<T = ExtractName<DefaultNS>> = T extends Fallback<string> ? T : string;
```
pedrodurek, line 174: Hey @JCQuintas, my suggestion would be something like this:
```ts
export interface CustomTypeOptions {}

type MergeBy<T, K> = Omit<T, keyof K> & K;
type TypeOptions = MergeBy<{
  defaultNS: 'translation';
}, CustomTypeOptions>

type DefaultNamespace<T = TypeOptions['defaultNS']> = T extends Fallback<string> ? T : string;
```

Then in the future, we could include resources here as well, even other options that may appear. So instead of having one type augmentation for each option, we could have one for all, for example:

```ts
interface CustomTypeOptions = {
  resources: typeof resources;
  defaulsNS: 'ns1';
}
```

What do you think?

JCQuintas: I like it. My approach was to not make changes to existing parts of the application cause I don't have a deep understanding of them. 

But I believe your approach to accomplish the same and still be flexible to future needs.

I'll make the changes

JCQuintas: I've changed it, incorporated the `resources` key as you suggested with backwards compatibility with the `Resources` type.

I've added tests to check if this works, though I had to update the typescript version. I don't think this will be an issue since it is a dev dep, but I don't know if it has other side effects I'm unaware of.

JCQuintas:
I've also taken the liberty to fix some `jest` warnings that appeared on the console by excluding the `/examples` folder

ts4.1/index.d.ts
pedrodurek: If we add resources now, we'll have breaking changes. I'm ok with having breaking changes, but we'd need to hold the PR off for a minor or major release.

//cc: @adrai

pedrodurek: Never mind, I've just seen you're keeping Resources and adding it to the default one. Just add a `@deprecated` decorator to the old `Resources` interface.

pedrodurek (approved):
LGTM 🎉 ! Thanks @JCQuintas! Just add the decorator as suggested and we can merge it!

ts4.1/index.d.ts
adrai: > If we add resources now, we'll have breaking changes. I'm ok with having breaking changes, but we'd need to hold the PR off for a minor or major release.

breaking means major

JCQuintas: Added `@deprecated` with a message to use `CustomTypeOptions` instead. And yes, it should be backwards compatible, so there would be no need for a `major` release

adrai:
so merge and minor version?

JCQuintas:
> so merge and minor version?

I believe so, yes

adrai:
@pedrodurek ok for you?

pedrodurek:
@adrai yeah, I'm ok with minor, thanks guys!

adrai:
Does something need to be changed here: https://react.i18next.com/latest/typescript ?

v11.11.0 is published

JCQuintas:
Documentation can be changed, should I make a PR for that as well? Any approach I should take?

adrai:
Feel free to create a PR

pedrodurek:
🙌

ts4.1/index.d.ts
ben-denzer: @JCQuintas This change did break my team's app. I haven't dug in deep enough to say if it is totally our fault or not - but I can say for sure that:

* Our app will not build with the latest version
* I went through about 6 build errors, hacking in some TS fixes, but the build errors just kept coming.
* I pin the version to "react-i18next": "11.8.6" and the app builds fine

ben-denzer: For anybody that makes it here with the same problem, we found an easy fix.

image

ben-denzer: For anybody who makes it here, we found an easy fix: 
image

PR #1341, opened 2021-07-13
feat(types): add returnNull and returnEptyString options to TypeOptions interface

ImADrafter:
Working with react-i18next and typescript, I found that **returnNull** and **returnEptyString** [options](https://www.i18next.com/overview/configuration-options#translation-defaults) were not represented in typescript signatures.

By providing keys for these values inside the `CustomTypeOptions`, we can refine the return types for TFunction. 

```tsx
const dictionary = {
    ...
    nullValue: null,
    emptyString: "",
}

t('nullValue') // By default, signature returns null.
t('emptyString') // By default, signature returns ""
```

```tsx
declare module 'react-i18next' {
    interface CustomTypeOptions extends MyCustomTypeOptions {
        returnNull: false,
        returnEmptyString: false,
    }
}

t('nullValue') // After moduleAugmentation, signature returns string, representing the key.
t('emptyString') // After moduleAugmentation, signature returns string, representing the key.
```

Notice that these keys must be aligned with the actual options in the i18next instance in order to represent the actual behaviour of the library on runtime. 

#### Checklist

- [x] tests are included

### DIFF
--- a/test/typescript/returnTypes.test.ts
+++ b/test/typescript/returnTypes.test.ts
@@ -1,0 +1,33 @@
+import { NormalizeByTypeOptions } from 'react-i18next';
+
+// Test cases for TypeOptions['returnNull']: true
+type ReturnNull = NormalizeByTypeOptions<null, { returnNull: true }>; // Returns null
+
+const nullableValue: ReturnNull = null;
+
+// @ts-expect-error: null is not assignable to string
+const nonNullableValue: ReturnNull = '';
+
+// Test cases for TypeOptions['returnNull']: false
+type ReturnNonNullable = NormalizeByTypeOptions<null, { returnNull: false }>; // Returns string
+
+// @ts-expect-error: null is not assignable to string
+const nullableValue2: ReturnNonNullable = null;
+
+const nonNullableValue2: ReturnNonNullable = '';
+
+// Test cases for TypeOptions['returnEmptyString']: false
+type ReturnNonEmptyString = NormalizeByTypeOptions<'', { returnEmptyString: false }>; // Returns string
+
+const emptyStringValue: ReturnNonEmptyString = '';
+
+// Emtpy string should always be assignable to string, but not viceversa
+const nonEmptyStringValue: ReturnNonEmptyString = 'non-empty-string';
+
+// Test cases for TypeOptions['returnEmptyString']: true
+type ReturnEmptyString = NormalizeByTypeOptions<'', { returnEmptyString: true }>; // Returns ""
+
+const emptyStringValue2: ReturnEmptyString = '';
+
+// @ts-expect-error: '"non-empty-string"' is not assignable to type '""'
+const nonEmptyStringValue2: ReturnEmptyString = 'non-empty-string';
--- a/ts4.1/index.d.ts
+++ b/ts4.1/index.d.ts
@@ -26,7 +26,7 @@ type Subtract<T extends K, K> = Omit<T, keyof K>;
  */
 export interface Resources {}
 /**
- * This interface can be augmented by users to add types to `react-i18next`. It accepts a `defaultNS` and `resources` properties.
+ * This interface can be augmented by users to add types to `react-i18next`. It accepts a `defaultNS`, `resources`, `returnNull` and `returnEmptyString` properties.
  *
  * Usage:
  * ```ts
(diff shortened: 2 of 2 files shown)


ts4.1/index.d.ts
```diff
@@ -111,12 +115,35 @@ type NormalizeMulti<T, U extends keyof T, L = LastOf<U>> = L extends U
...
+  U extends Partial<CustomTypeParameters>
+> = U['returnEmptyString'] extends false ? (T extends '' ? string : T) : T;
+
+/**
+ * Checks if user has enabled `returnEmptyString` and `returnNull` options to retrieve correct values.
+ */
+export type NormalizeByTypeOptions<
+  T,
+  U extends Partial<CustomTypeParameters> = TypeOptions
+> = NonNullableTranslation<NonEmptyStringTranslation<T, U>, U>;
```
pedrodurek, line 139: As both helpers are pretty much the same, you could make it more reusable, something like:
```suggestion
type TypeOptionsFallback<T, U, R> = U extends false ? (T extends R ? string : T) : T

/**
 * Checks if user has enabled `returnEmptyString` and `returnNull` options to retrieve correct values.
 */
export type NormalizeByTypeOptions<
  T,
  U extends Partial<CustomTypeParameters> = TypeOptions,
  R = TypeOptionsFallback<T, U['returnEmptyString'], ''>
> = TypeOptionsFallback<R, U['returnNull'], null>;
```

ts4.1/index.d.ts
```diff
@@ -111,12 +115,35 @@ type NormalizeMulti<T, U extends keyof T, L = LastOf<U>> = L extends U
   ? AppendNS<L, Normalize<T[L]>> | NormalizeMulti<T, Exclude<U, L>>
   : never;
 
+type CustomTypeParameters = {
+  returnNull: boolean;
+  returnEmptyString: boolean;
+};
```
pedrodurek, line 121: Can't we simply rely on `type CustomTypeParameters = Pick<TypeOptions, 'returnNull' | 'returnEmptyString'>;`?

In case we can't, we can make both parameters optional instead of using partial in three different places:
```ts
type CustomTypeParameters = {
  returnNull?: boolean;
  returnEmptyString?: boolean;
};
```

test/typescript/returnTypes.test.ts
```diff
@@ -0,0 +1,33 @@
...
+// Emtpy string should always be assignable to string, but not viceversa
+const nonEmptyStringValue: ReturnNonEmptyString = 'non-empty-string';
+
+// Test cases for TypeOptions['returnEmptyString']: true
+type ReturnEmptyString = NormalizeByTypeOptions<'', { returnEmptyString: true }>; // Returns ""
+
+const emptyStringValue2: ReturnEmptyString = '';
+
+// @ts-expect-error: '"non-empty-string"' is not assignable to type '""'
+const nonEmptyStringValue2: ReturnEmptyString = 'non-empty-string';
```
pedrodurek, line 33: I'm always in favour of testing everything integrated by testing both flags in the `useTranslation`, but I'm afraid it's not possible to toggle them on/off 🤔

ts4.1/index.d.ts
ImADrafter: Yup! I applied your suggestion, but changing the generic parameters naming. What do you think ? @pedrodurek

ts4.1/index.d.ts
ImADrafter: We can't make that Pick, because the types for returnNull and returnEmptyString would be `true`, not boolean :(

Applied your suggestion about making the keys optional and remove that Partial @pedrodurek

test/typescript/returnTypes.test.ts
ImADrafter: Yep, I don't know if there's a way to make the augmentation twice to check both flags. If you have any idea, just let me know :) @pedrodurek

ts4.1/index.d.ts
pedrodurek: Sounds good to me!

adrai:
is this ready to merge?

pedrodurek (approved):
LGTM! 🎉

jamuhl:
published in react-i18next@11.11.4

Issue #1344, opened 2021-07-15
Trans component „closes“ interpolation values in `<>` when `escapeValue: false` is set


ckruse:
## 🐛 Bug Report

When interpolating a value containing something remotely similar to an HTML tag `Trans` „closes“ that tag. For example:

~~~js
const email = "Foo <bar@example.com>";
return <Trans parent="div"><span>email:</span> {{email}}</Trans>;
~~~

This generates this output:

~~~html
<div><span>email:</span> Foo &lt;bar@example.com&gt;</div>
~~~

It doesn't matter if I enable or disable `transSupportBasicHtmlNodes`.

## To Reproduce

- https://codesandbox.io/s/peaceful-sound-2xmrz
- https://codesandbox.io/s/purple-wood-0cyef

## Expected behavior

The interpolation value w/o any changes should be displayed

## Your Environment

- *runtime version*: Browser
- *i18next version*: i18next@20.3.3, react-i18next@11.11.2
- *os*: Mac, Linux

adrai:
This is probably because of html-parse-stringify...
Open to accept a PR: https://github.com/i18next/react-i18next/blob/master/src/Trans.js#L138

ckruse:
Ok, I understand why interpolation happens before parsing, but to me it seems that the only possible fix to this problem is to do it the other way round. Do you agree or am I wrong?

adrai:
Currently afk, so I did not investigate this issue.
Feel free to suggest a fix.
It will then be reviewed by me or @jamuhl

jamuhl:
just a reminder for the review the reason why we added interpolation upfront: https://github.com/i18next/react-i18next/blame/v9.x.x/src/Trans.js#L61

with a first look at the PR this seems to still work

ckruse:
Fixed by PR

adrai:
releasd in v11.11.3

PR #1345, opened 2021-07-15
parse first, then interpolate

ckruse:
This avoids problems with user given values potentially breaking the parse tree

Regarding https://github.com/i18next/react-i18next/issues/1344

#### Checklist

- [x] tests are included

### DIFF
--- a/src/Trans.js
+++ b/src/Trans.js
@@ -127,15 +127,10 @@ function renderNodes(children, targetString, i18n, i18nOptions, combinedTOpts) {
 
   getData(children);
 
-  const interpolatedString = i18n.services.interpolator.interpolate(
-    targetString,
-    { ...data, ...combinedTOpts },
-    i18n.language,
-  );
-
   // parse ast from string with additional wrapper tag
   // -> avoids issues in parser removing prepending text nodes
-  const ast = HTML.parse(`<0>${interpolatedString}</0>`);
+  const ast = HTML.parse(`<0>${targetString}</0>`);
+  const opts = { ...data, ...combinedTOpts };
 
   function renderInner(child, node, rootReactNode) {
     const childs = getChildren(child);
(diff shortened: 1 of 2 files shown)


adrai:
lgtm

adrai:
should be still backwards compatible, right?
So patch or minor version is sufficient?

jamuhl:
lgtm.... https://github.com/i18next/react-i18next/blob/master/test/trans.render.spec.js#L336 -> initial reason should be covered by tests

adrai:
included in v11.11.3

adrai:
Thank you @ckruse for your contribution.

ckruse:
Thank you for your work!

dlavrenuek:
I wanted to create an issue but since this was fixed in this PR I would like to comment with some additional information. The `Trans` component in `react-i18next` < 11.11.3 is vulnerable to XSS attacks because the provided variables were rendered as HTML. Example code:

```javascript

function App() {
    const [text, setText] = useState('');

    return (
        <>
            <input
                value={text}
                onChange={(e) => setText(e.target.value)}
            />
            <p><Trans>You entered {{text}}</Trans></p>
        </>
    );
}
```

dlavrenuek:
@jamuhl @adrai According to https://nodejs.org/en/security/#reporting-a-bug-in-a-third-party-module a package owner should create a security report, so the vulnerability is properly reported by npm vulnerability check and tools like dependabot.

adrai:
@dlavrenuek 
with react-i18next v11.0.0 - v11.11.4 it's always the same:
input: `my <b>variable</b> <div>and other</div>`
output: `<p>You entered my &lt;b&gt;variable&lt;/b&gt; &lt;div&gt;and other&lt;/div&gt;</p>`

Maybe you compared it to a version < v10.12.2 ? => https://github.com/i18next/react-i18next/blob/master/CHANGELOG.md#10122

dlavrenuek:
@adrai I missed to provide that `Trans` was only vulnerable to XSS when setting the option `escapeValue` to `true` which is suggested as default by https://react.i18next.com/guides/quick-start#configure-i-18-next

```
interpolation: {
    escapeValue: true
}
```

So far I was not successful at executing javascript with it due to restrictions with `transKeepBasicHtmlNodesFor`, but you can destroy the page layout by adding a lot of `<br />` and `<p>` and may be exploit it in some way.

jamuhl:
`escapeValue` is suggested to be set to false -> as react already escapes strings rendered in elements.

this was tested long ago and there is yet no known way of injecting executable code into `Trans` - but we're happy to be proven wrong

As the `transKeepBasicHtmlNodesFor` come from your translation I guess that is save

PR #1368, opened 2021-09-01
Fix preact issue with Trans when both children and components are undefined

visualfanatic:
### Description

I stumbled upon a bug when integrating `react-i18next` in an app that uses preact. It turns out that if you use `Trans` component with no `children` and you do not pass any `components`, the whole translation will not render. It works with React, but not with Preact. Here are simple reproductions:

- React: https://codesandbox.io/s/react-3txsi
- Preact (no `children` and no `components`): https://codesandbox.io/s/preact-trans-no-children-no-components-unwxw
- Preact (no `children`, empty object passed to `components`): https://codesandbox.io/s/preact-trans-no-children-empty-object-components-wnrmq

#### Checklist

- [x] tests are included

### DIFF
--- a/src/Trans.js
+++ b/src/Trans.js
@@ -258,7 +258,11 @@ function renderNodes(children, targetString, i18n, i18nOptions, combinedTOpts) {
   // call mapAST with having react nodes nested into additional node like
   // we did for the string ast from translation
   // return the children of that extra node to get expected result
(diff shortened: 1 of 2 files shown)


adrai:
lgtm

jamuhl (approved):
lgtm

adrai:
thank you for your contribution, this will land in the next version.... probably in the next days...

PR #1374, opened 2021-09-15
fix: update import

maapteh:
Correct example with latest setup


jamuhl:
that's a sample of v9 -> https://react.i18next.com/legacy-v9/i18next-instance so the change is wrong

maapteh:
No clue how i ended up from new docs into the old ones :(

PR #1426, opened 2021-12-07
fix: unescape html in Trans

### LINKED ISSUE #1424: Trans should allow html encoded character like `&lt; &amp; &gt;`
## 🐛 Bug Report

[Trans](https://react.i18next.com/latest/trans-component) should allow html encoded character like `&lt; &amp; &gt;`

## To Reproduce

[https://codesandbox.io/s/exciting-wescoff-kjz0j](https://codesandbox.io/s/exciting-wescoff-kjz0j)
## Expected behavior

Render `< & >` correctly.

Ref: https://github.com/i18next/i18next/issues/1567

## Your Environment

- *runtime version*:  node v16
- *i18next version*:  21.0.0
- *react-i18next version*: 11.10.0
- *os*: Mac, Windows, Linux

cuyl:
fixes #1424


jamuhl:
Not sure if this should be an optional behaviour - can't really guess if there are side effects on existing projects? @adrai what do you think?

cuyl:
The content should escape when extracting Trans content, but I do not use *saveMissing*.

It may be a breaking change unless you can share translation memory between escaped and unescaped versions.

adrai:
my gut feeling says, this should only be introduced with a new Trans prop that enables this

adrai:
@cuyl can you adapt the PR accordingly?

cuyl:
@adrai Sure, please check it out

adrai:
@jamuhl lgtm (maybe also a react init option?)

jamuhl:
@adrai lgtm...guess react init option could be added in a later stage if really used that often?

jamuhl:
published in react-i18next@11.15.0

Issue #1459, opened 2022-03-01
Type error with typescript 4.6


jansepke:
## 🐛 Bug Report

After upgrading typescript from 4.5.5 to 4.6.2 we get the following error when running tsc:

```
node_modules/react-i18next/ts4.1/index.d.ts:333:14 - error TS2344: Type 'ComponentProps<C>' does not satisfy the constraint 'WithTranslationProps'.
  Type 'unknown' is not assignable to type 'WithTranslationProps'.

333     Subtract<React.ComponentProps<C>, WithTranslationProps>
                 ~~~~~~~~~~~~~~~~~~~~~~~


Found 1 error in node_modules/react-i18next/ts4.1/index.d.ts:333
```

## Expected behavior

react-i18next should also be valid typescript for typescript 4.6

## Your Environment

- *runtime version*: node v14.19.0
- *i18next version*: 21.6.0
- *os*: Linux

cojoclaudiu:
I confirm the same error Vite React Typescript

```
> tsc && vite build

node_modules/react-i18next/ts4.1/index.d.ts:333:14 - error TS2344: Type 'ComponentProps<C>' does not satisfy the constraint 'WithTranslationProps'.
  Type 'unknown' is not assignable to type 'WithTranslationProps'.

333     Subtract<React.ComponentProps<C>, WithTranslationProps>
                 ~~~~~~~~~~~~~~~~~~~~~~~


Found 1 error in node_modules/react-i18next/ts4.1/index.d.ts:333
```

Anubiso:
Having the same issue

wizzymotion:
same issue here using VITE too

ortegarenzy:
Same issue:

Vite 2.8.0
TypeScript 4.5.5
ViteJS Plugin React 1.0.7

```
node_modules/react-i18next/ts4.1/index.d.ts:333:14 - error TS2344: Type 'ComponentProps<C>' does not satisfy the constraint 'WithTranslationProps'.
  Type 'unknown' is not assignable to type 'WithTranslationProps'.

333     Subtract<React.ComponentProps<C>, WithTranslationProps>
                 ~~~~~~~~~~~~~~~~~~~~~~~


Found 1 error in node_modules/react-i18next/ts4.1/index.d.ts:333

error Command failed with exit code 2.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

cojoclaudiu:
> TypeScript 4.5.5
[...]

PR #1463, opened 2022-03-07
Fix error for typescript 4.6

pedrodurek:
Closes https://github.com/i18next/react-i18next/issues/1459 and bump ts version

#### Checklist

- [x] tests are included

### DIFF
--- a/ts4.1/index.d.ts
+++ b/ts4.1/index.d.ts
@@ -327,7 +327,7 @@ export function withTranslation<N extends Namespace = DefaultNamespace>(
     withRef?: boolean;
   },
 ): <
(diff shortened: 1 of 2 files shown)


ts4.1/index.d.ts
```diff
@@ -327,7 +327,7 @@ export function withTranslation<N extends Namespace = DefaultNamespace>(
     withRef?: boolean;
   },
 ): <
-  C extends React.ComponentType<React.ComponentProps<C> & WithTranslationProps>,
```
pedrodurek, line 330: The constrain should not rely on its on generics, this breaks on typescript 4.6 as `C` becomes `unknown` rather than any in the first instance

adrai:
included in v11.15.6

PR #1472, opened 2022-03-24
Update macro to wrap defaults in brackets when necessary

heyimalex:
I noticed that the Trans macro generates invalid code when children contains a double quote. From [ast explorer](https://astexplorer.net/#/gist/642aebbb9e449e959f4ad8907b4adf3a/4a65742e2a3e926eb55eaa3d657d1472b9ac7970):

```jsx
// before
<Trans>"</Trans>;
// after
<Trans defaults="\"" components={[]} values={{}} />;
```

According to the [JSX spec](https://facebook.github.io/jsx/#prod-JSXAttributeValue) this isn't valid; JSX doesn't support any kind of escaping inside of `"` quoted attributes. Babel, tsc, prettier, etc all fail when parsing. To get around this we can just wrap the value in brackets. I made it only happen when the string in question contains a quote.

```diff
// before
<Trans>"</Trans>;
// after
- <Trans defaults="\"" components={[]} values={{}} />;
+ <Trans defaults={"\""} components={[]} values={{}} />;
```

I am kinda confused about how this is working today without issue.

I think we could also `defaults.replace(/"/g, "&quot;")` instead if that's preferred.

#### Checklist

- [x] tests are included

### DIFF
--- a/icu.macro.js
+++ b/icu.macro.js
@@ -263,7 +263,10 @@ function buildTransElement(
   // add generated Trans attributes
   if (!attributeExistsAlready('defaults', finalAttributes))
     finalAttributes.push(
-      t.jSXAttribute(t.jSXIdentifier('defaults'), t.StringLiteral(extracted.defaults)),
+      t.jSXAttribute(
+        t.jSXIdentifier('defaults'),
+        t.jSXExpressionContainer(t.StringLiteral(extracted.defaults)),
+      ),
     );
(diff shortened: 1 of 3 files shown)


jamuhl (approved):
Don't know the specs in detail but assumed we can just escape...guess this is ok...

adrai:
It's included in v11.16.2

PR #1480, opened 2022-04-10
Fix type 'TFunctionResult' is not assignable to type 'ReactNode' on React 18

akunzai:
Fixes #1476 by add `ReactNode` to the `TDefaultResult` type

#### Checklist

- [x] only relevant code is changed (make a diff before you submit the PR)
- [x] run tests `npm run test`
- [ ] tests are included

PR #1481, opened 2022-04-10
ReactNode should be prefixed with React

high1:
ReactNode was incorrectly added in the #1480 without React prefix - breaks tsc compilation, unless skipLibCheck is set.

#### Checklist

- [ ] tests are included


adrai:
thx, included in v11.16.5

high1:
@adrai The tests are failing because of tslint - TFunctionResult that's imported from i18next is defined as a union of both null and undefined, and that's an error.

adrai:
maybe @pedrodurek can assist here

PR #1482, opened 2022-04-10
Disable no-null-undefined-union for TFunctionResult

high1:
Disable no-null-undefined-union rule for TFunctionResult imported from i18next.

#### Checklist

- [ ] tests are included


adrai:
thx @high1, does this need to be published in a new npm version?

high1:
@adrai - I don't think it does, as linting is normally done only on project source, not dependencies, so it can be included in the next release, when due.

Issue #1490, opened 2022-04-27
React 18 - Trans component interpolation issue


enoh-barbu:
I've just upgraded to react v18 but I get an error while doing interpolation using Trans component

```
<Trans i18nKey="pageNrOfAll" pageNr={pageNr} pagesCount={pagesCount}>
    Page
    <strong className="spacing">{{ pageNr }}</strong>
    of
    <strong className="spacing">{{ pagesCount }}</strong>
 </Trans>
```
![image](https://user-images.githubusercontent.com/24221572/165543530-02cc0e7e-cdc9-446e-b001-f9814c51302e.png)

```
Type '{ pageNr: number; }' is not assignable to type 'ReactNode'.
  Object literal may only specify known properties, and 'pageNr' does not exist in type 'ReactElement<any, string | JSXElementConstructor<any>> | ReactFragment | ReactPortal'.ts(2322)
index.d.ts(1375, 9): The expected type comes from property 'children' which is declared here on type 'DetailedHTMLProps<HTMLAttributes<HTMLElement>, HTMLElement>'
```
If I keep only one pair of brackets `{ pageNr }` I get the following error

`react-i18next:: Trans: the passed in value is invalid - seems you passed in a variable like {number} - please pass in variables for interpolation as full objects like {{number}}.`

[https://codesandbox.io/s/brave-bohr-obwf63](https://codesandbox.io/s/brave-bohr-obwf63)

I've just posted the same question on [stackoverflow](https://stackoverflow.com/questions/72030446/react-18-react-i18next-trans-component-interpolation-issue)

adrai:
probably duplicate of #1483

adrai:
v11.16.8 offers allowObjectInHTMLChildren option

Issue #1483, opened 2022-04-11
Cannot pass objects to `<Trans>` with React 18 type definitions


LukasKalbertodt:
## 🐛 Bug Report

With version 18 of `@types/react`, Typescript complains when passing objects to the `Trans` component. 

It previously worked because `ReactNode` had a `| {}` in its definition, allowing arbitrary objects to be passed. That has been removed (for a good reason) in version 18 of the type definitions. See [this PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/56210) for more information.

## To Reproduce

From [this example](https://react.i18next.com/latest/trans-component#using-with-react-components):

```jsx
<Trans i18nKey="userMessagesUnread" count={count}>
  Hello <strong title={t('nameTitle')}>{{name}}</strong>, you have {{count}} unread message. <Link to="/msgs">Go to messages</Link>.
</Trans>
```

This fails type checking: `TS2322: Type '{ name: void; }' is not assignable to type 'ReactNode'.` 

## Expected behavior

Type checking works, i.e. the TS compiler does not complain.

I would think that `{}` (or something like `Record<string, unknown>`) should be explicitly added to the `children` type of `Trans`.

## Your Environment

- *react version*: 18.0.0
- *i18next version*: 11.16.5
- *TS version*: 4.6.3

adrai:
@high1 maybe related?

high1:
@adrai - React did remove implicit children on their components in @types/react18, as @LukasKalbertodt aleady said, so I would expect something like this to happen - if this is children related, Trans component just need to allow children explicitly.
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/56210

high1:
I took a look at this today. @LukasKalbertodt is 100% on the point - another breaking change is that objects are removed from ReactNode. And the suggested solution should work - but if React.ReactNode | Record<string, unknown> gets added to defintions, another null-undefined union error happens. I would disable this rule at this point, if that's fine, @adrai - and update tslint to it's latest version, 6.1.3. And then I would suggest that the repo upgrades to @typescript-eslint.

adrai:
need @pedrodurek's feedback here 🙏

LukasKalbertodt:
This is not yet fixed. The code in the top level comment now (11.16.7) results in:

```
ERROR 
[...]

PR #1492, opened 2022-04-28
Fix Trans component to support react 18 types

pedrodurek:
Closes https://github.com/i18next/react-i18next/issues/1490 and https://github.com/i18next/react-i18next/issues/1483

#### Checklist

- [ ] tests are included

### DIFF
--- a/ts4.1/index.d.ts
+++ b/ts4.1/index.d.ts
@@ -40,6 +40,7 @@ export interface Resources {}
  *     nsSeparator: ':';
  *     keySeparator: '.';
  *     jsonFormat: 'v4';
+ *     allowObjectInHTMLChildren: false;
  *     resources: {
  *       custom: {
  *         foo: 'foo';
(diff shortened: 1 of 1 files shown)


adrai:
included in v11.16.8

ts4.1/index.d.ts
```diff
@@ -97,6 +99,15 @@ declare module 'i18next' {
...
 }
 
+type ObjectOrNever = TypeOptions['allowObjectInHTMLChildren'] extends true
+  ? Record<string, unknown>
+  : never;
+declare module 'react' {
+  interface HTMLAttributes {
+    children?: ReactNode | ObjectOrNever;
+  }
+}
```
cellog, line 109: this override is breaking all components that use render props when upgrading to React 18. We can patch this away, but I would strongly encourage NOT overriding generic types for all components. Can't it be limited to just the `Trans` component? A sledgehammer can't be the only solution when a scalpel will do?

adrai: Feel free to provide a PR with a better solution that works.

// @marcalexiei do you know about this?

pedrodurek: @cellog **extending** is different from **overriding**. Share a minimal reproducible example so we can see how we can help you.

PR #1493, opened 2022-04-28
Fix "order of Hooks" warning if i18n doesn't exist yet

ymoran00:
If useTranslation() is called before i18n initializes (which can happen - I called `useTranslation()` to get the isReady parameter to know the state of i18n) we get a warning in the console:
`Warning: React has detected a change in the order of Hooks called by...`

In this PR I only changed the order of the commands in useTranslation() to fix this issue - having the useState and useRef happen before we exit the function when there's no i18n.

#### Checklist

- [x] tests are included

### DIFF
--- a/src/useTranslation.js
+++ b/src/useTranslation.js
@@ -8,6 +8,33 @@ export function useTranslation(ns, props = {}) {
   const { i18n: i18nFromContext, defaultNS: defaultNSFromContext } = useContext(I18nContext) || {};
   const i18n = i18nFromProps || i18nFromContext || getI18n();
   if (i18n && !i18n.reportNamespaces) i18n.reportNamespaces = new ReportNamespaces();
+
+  if (i18n && i18n.options.react && i18n.options.react.wait !== undefined)
+    warnOnce(
+      'It seems you are still using the old wait option, you may migrate to the new useSuspense behaviour.',
+    );
+
+  const i18nOptions = i18n ? { ...getDefaults(), ...i18n.options.react, ...props } : getDefaults();
+  const { useSuspense, keyPrefix } = i18nOptions;
+
+  // prepare having a namespace
+  let namespaces = ns || defaultNSFromContext || (i18n && i18n.options && i18n.options.defaultNS);
+  namespaces = typeof namespaces === 'string' ? [namespaces] : namespaces || ['translation'];
+
+  // binding t function to namespace (acts also as rerender trigger)
+  function getT() {
+    return (
+      i18n &&
+      i18n.getFixedT(
+        null,
+        i18nOptions.nsMode === 'fallback' ? namespaces : namespaces[0],
+        keyPrefix,
+      )
+    );
+  }
+  const [t, setT] = useState(getT);
+
+  const isMounted = useRef(true);
   if (!i18n) {
     warnOnce('You will need to pass in an i18next instance by using initReactI18next');
     const notReadyT = (k) => (Array.isArray(k) ? k[k.length - 1] : k);
@@ -18,18 +45,6 @@ export function useTranslation(ns, props = {}) {
     return retNotReady;
   }
 
-  if (i18n.options.react && i18n.options.react.wait !== undefined)
-    warnOnce(
-      'It seems you are still using the old wait option, you may migrate to the new useSuspense behaviour.',
-    );
-
-  const i18nOptions = { ...getDefaults(), ...i18n.options.react, ...props };
-  const { useSuspense, keyPrefix } = i18nOptions;
-
-  // prepare having a namespace
-  let namespaces = ns || defaultNSFromContext || (i18n.options && i18n.options.defaultNS);
-  namespaces = typeof namespaces === 'string' ? [namespaces] : namespaces || ['translation'];
-
   // report namespaces as used
   if (i18n.reportNamespaces.addUsedNamespaces) i18n.reportNamespaces.addUsedNamespaces(namespaces);
 
@@ -38,17 +53,6 @@ export function useTranslation(ns, props = {}) {
     (i18n.isInitialized || i18n.initializedStoreOnce) &&
     namespaces.every((n) => hasLoadedNamespace(n, i18n, i18nOptions));
 
-  // binding t function to namespace (acts also as rerender trigger)
-  function getT() {
-    return i18n.getFixedT(
-      null,
-      i18nOptions.nsMode === 'fallback' ? namespaces : namespaces[0],
-      keyPrefix,
-    );
-  }
-  const [t, setT] = useState(getT);
-
-  const isMounted = useRef(true);
   useEffect(() => {
     const { bindI18n, bindI18nStore } = i18nOptions;
     isMounted.current = true;


jamuhl:
@adrai  I'm for a simpler change...instead of warn if i18n is not defined - just throw an error

jamuhl:
I'm not sure why it's important to follow any "rules of hooks" if the app has an error that needs to be fixed

jamuhl:
It's not that it's hard to fix the underlaying issue - it's not that it requests you to have translations loaded or an async init done...it just asks for having a i18n instance to do those checks for you

adrai:
> @adrai I'm for a simpler change...instead of warn if i18n is not defined - just throw an error

@ymoran00 would throwing an error be ok for you?

ymoran00:
No, because this is not an error.
I have a component that runs while the app is loading - a <Loading/> component. If we don't have i18n yet it should display the text in English, otherwise - translated. So it may appear whill app loads and we don't have i18n ye - and it's ok. We don't want an error in this case.

adrai:
Can you please provide a reproducible example showing your use case... sounds strange... there should aways be an i18next instance...

jamuhl:
There are better solutions to solve this.

```js
import React from 'react';
import { Markdown } from '@locize/locize-visual';
import { useTranslation } from '../utils';

export function Text({
  i18nKey,
  ns = 'client-locize-app-loaders',
  i18nOptions = {},
  message,
  defaultValue,
  linkTarget = '_blank',
}) {
  const { t, ready } = useTranslation(ns, { useSuspense: false });
  if (!ready)
    return (
      <Markdown source={message || 'loading...'} noContainer options={{ linkTarget: '_blank' }} />
    );

  return (
    <Markdown
      source={message || t(i18nKey, { defaultValue, ...i18nOptions })}
      noContainer
      options={{ linkTarget }}
    />
  );
}

```

Not having an i18n instance is an error per definition of it not working without. Another issue I don't understand how can't you have an i18n instance. I understand translations might not yet be loaded - but that can solved like shown above (for those loading things)

ymoran00:
> Can you please provide a reproducible example showing your use case... sounds strange... there should aways be an i18next instance...

In my case I know the language of the user only after receiving it from server. So if I init the i18n instance first thing in the app (as the documentation suggests), it means I'll need to later changeLanguage in i18n. This causes multuiple language fetches (for English and for the user's language) and a flicker in the UI.
So in my code I delayed initializing the i18n instance - and this caused this problem.
Am I doing it wrong?

adrai:
> > Can you please provide a reproducible example showing your use case... sounds strange... there should aways be an i18next instance...
> 
> In my case I know the language of the user only after receiving it from server. So if I init the i18n instance first thing in the app (as the documentation suggests), it means I'll need to later changeLanguage in i18n. This causes multuiple language fetches (for English and for the user's language) and a flicker in the UI. So in my code I delayed initializing the i18n instance - and this caused this problem. Am I doing it wrong?

The question is, what are you doing until you fetched the persisted user language? Do you show some fallback UI without any text? Do you use the fallbackLng? Do you use the detected browser language?
If you're just showing a loading spinner or similar, just do not call/render any t() invocations... so basically do not use i18next before you have loaded the persisted user language...
Personally, I would show a fallback language or at least a detected browser language... (fyi: you can also create your own language detector: https://stackoverflow.com/a/59893089/1859027)
So to come back to the most important question: "What do you want to do/show until you know the persisted user language?"

ymoran00:
This is what I show while waiting:

```
export const Loading = (): JSX.Element => {
    const {t, ready} = useTranslation();
    return (
        <Wrapper>
            <Message>
                {ready ? t('components.pages.loading', 'Loading') : 'Loading'}
            </Message>
            <Progress/>
        </Wrapper>
    )
}
```
I want to have a spinner and a "Loading..." text. Since the same component is used throughout the app whenever an async request is ongoing, I use i18n to translate the "Loading". But before the first initialization I have the issues we've difscussed.

adrai:
So you have to choose...

1. show always the hardcoded Loading text (or `t('key', fallback text')`)
2. do not show any text while loading
3. live with a possible text change as soon as ready

@jamuhl do I miss something?

jamuhl:
no. it‘s what we do using my posted snipplet

adrai:
I suggest, in the next major, we'll throw if no i18n instance is provided

ymoran00:
> So you have to choose...
> 
> 1. show always the hardcoded Loading text (or `t('key', fallback text')`)
> 2. do not show any text while loading
> 3. live with a possible text change as soon as ready
> 
> @jamuhl do I miss something?

So I'll add 4 - use a different component for loading the entire app with an English "Loading", and have all other "Loading" in app use translation.
With this I can live.

adrai:
if you set the fallbackLng option to english, it is exactly that

jamuhl:
@ymoran00 that's what my pasted component does...if i18next is ready use translation - else the "hardcoded" english fallback...but as you got your solution I think we can close this now

ymoran00:
> if you set the fallbackLng option to english, it is exactly that

It isn't - this options is when initializing the i18n instance, but as I explained here - I can't initialize it before the app loads and I know what language to initialize with, so I'm stuck with this warning. It seems I need a different component without usage of i18n at all while loading the app.

adrai:
Can't you initialize i18next with your fallbackLng (English) only? Why not?

I think we're missing something you're trying to explain...
Best would be you create a reproducible example, so we exactly see what you mean.

ymoran00:
> @ymoran00 that's what my pasted component does...if i18next is ready use translation - else the "hardcoded" english fallback...but as you got your solution I think we can close this now

Your pasted component has the same proble - it uses `useTranslation` to check if it's ready - but then the warning appears because it needs the i18n to be ready... This is a loop...

jamuhl:
it does not need i18n to be ready...it just needs i18n instance...initialized, loaded translations or not...irrelevant...it just needs an instance providing expected API...as it checks if i18n is initialized and ready.

and having an i18n instance passed `useTranslation` works with or without fallback lng...just not without lng...and that is your problem I guess...you do not have a language and you do not like to show a fallback at all until you get the language from your backend API.

Could be solved either by using a custom language detector having async true (As you call some async request) or like you do...just have a simple component avoid loading at all as the app depends on your backends language.

adrai:
fyi: here an example with a custom language detector: https://codesandbox.io/s/react-i18next-http-example-forked-h15xb5?file=/src/i18n.js

jamuhl:
Just checked...one of the problems is that when using `initReactI18next` by `i18n.use` --> that runs on init so `useTranslation` has no i18n instance in that case...

Simple workaround is using `setI18n`:  https://codesandbox.io/s/react-i18next-forked-s2z0pr?file=/src/i18n.js

=> That works triggering the not ready Suspense and init async at a later point

ymoran00:
> it does not need i18n to be ready...it just needs i18n instance...initialized, loaded translations or not...irrelevant...it just needs an instance providing expected API...as it checks if i18n is initialized and ready.
> 
> and having an i18n instance passed `useTranslation` works with or without fallback lng...just not without lng (but that's more a problem of i18next itself that can't be initialized without lng or detector - i guess)...and that is your problem I guess...you do not have a language and you do not like to show a fallback at all until you get the language from your backend API.
> 
> Could be solved either by using a custom language detector having async true https://www.i18next.com/misc/creating-own-plugins#languagedetector (As you call some async request) or like you do...just have a simple component avoid loading at all as the app depends on your backends language.

Here's an example I've created. Added a timeout before loading i18n to simulate waiting fro langauge from server:
https://codesandbox.io/s/react-i18n-loading-warning-0q2d7o

ymoran00:
> fyi: here an example with a custom language detector: https://codesandbox.io/s/react-i18next-http-example-forked-h15xb5?file=/src/i18n.js

Ok, this may have worked - I've user setI18n() but also had to set the default of i18n.options.react.useSuspense to false:

https://codesandbox.io/s/react-i18n-loading-warning-0q2d7o?file=/src/index.tsx

PR #1529, opened 2022-07-03
feat: enable html preprocessing

### LINKED ISSUE #1435: Unescape soft-hypen
## 🐛 Bug Report

The `<Trans/>` component has the property `shouldUnescape` to allow unescaping of HTML entities like `&lt;` or `&amp;` to become `<` and `&` respectively. However, this mechanism only works for a subset of HTML entities, in particular it does not work for soft-hyphens: `&shy;` won't turn into the invisible soft hyphen character (`U+00AD`).

## To Reproduce

Here's a [codesanbox](https://codesandbox.io/s/react-i18next-forked-2j7o6) showing the issue. You can also reproduce this by using the `<Trans/>` component as follows with a translation string that includes `&shy`: `<Trans shouldUnescape={true}>`

## Expected behavior

`&shy;` should be unescaped into the invisible soft hyphen character (`U+00AD`).

## Your Environment

- *runtime version*: Node 14, Chrome/FF
- *i18next version*: 11.15.1
- *os*: Windows, Linux

nilsmehlhorn:
#### Checklist

- [x] tests are included

Closes #1435

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -44,7 +44,7 @@ export interface TransProps<E extends Element = HTMLDivElement>
   parent?: string | React.ComponentType<any> | null; // used in React.createElement if not null
   tOptions?: {};
   values?: {};
-  shouldUnescape?: boolean;
+  preprocessor?: (html: string) => string;
   t?: TFunction;
 }
 export function Trans<E extends Element = HTMLDivElement>(props: TransProps<E>): React.ReactElement;
--- a/index.d.ts
+++ b/index.d.ts
@@ -44,7 +44,7 @@ export interface TransProps<E extends Element = HTMLDivElement>
   parent?: string | React.ComponentType<any> | null; // used in React.createElement if not null
   tOptions?: {};
   values?: {};
-  preprocessor?: (html: string) => string;
+  shouldUnescape?: boolean;
   t?: TFunction;
 }
 export function Trans<E extends Element = HTMLDivElement>(props: TransProps<E>): React.ReactElement;
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,6 @@
       "license": "MIT",
       "dependencies": {
         "@babel/runtime": "^7.14.5",
-        "html-escaper": "^2.0.2",
         "html-parse-stringify": "^3.0.1"
       },
       "devDependencies": {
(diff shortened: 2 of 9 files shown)


adrai:
probably major version, right?

nilsmehlhorn:
> probably major version, right?

Breaking change in the component interface so I'd say yes

jamuhl:
Is there a specific reason to remove the existing functionality to unescape and making this a breaking version.

My suggestion would be:

1) include the html-escaper function (copy&paste - not sure why this was included in first place as the function is too small to justify adding it as a dependency)
2) let devs override that unescape function

Also I would keep the `unescape` - naming it preprocessor makes it sound like it gets applied to all content - while it only gets applied on text nodes.

adrai:
fyi: https://github.com/i18next/react-i18next/pull/1426

nilsmehlhorn:
> include the html-escaper function (copy&paste - not sure why this was included in first place as the function is too small to justify adding it as a dependency) => https://github.com/WebReflection/html-escaper/blob/master/index.js

That entity subset is actually too small for me and others (see #1435). There's a bunch more entities, e.g. visible in alt dependency: https://github.com/mdevils/html-entities/blob/master/src/named-references.ts

> let devs override that unescape function

So, like this?

```ts
type TransProps =  {
  ...
  shouldUnescape?: boolean | (text: string) => string;
}
```

> Also I would keep the `unescape` - naming it preprocessor makes it sound like it gets applied to all content - while it only gets applied on text nodes.

Naming is actually `shouldUnescape`, which doesn't necessarily work well for function but idk. Could also go with `textPreprocessor`.

nilsmehlhorn:
I've implemented a default unescape analogue to html-escaper and enabled the override via options. Not quite sure about the typing though, because `ReactOptions` seems to come from the i18n package itself which doesn't really add up for me - do we need a change in that package to update the options type?

adrai:
> I've implemented a default unescape analogue to html-escaper and enabled the override via options. Not quite sure about the typing though, because `ReactOptions` seems to come from the i18n package itself which doesn't really add up for me - do we need a change in that package to update the options type?

Yes, currently the types needs to be updated here for react-i18next: https://github.com/i18next/i18next/blob/master/index.d.ts#L186 correct me @pedrodurek if I'm wrong

adrai:
released with react-i18next v11.18.0 and i18next v21.8.13

adrai:
@nilsmehlhorn thank you for your contribution

PR #1531, opened 2022-07-05
allow iterable with objects as children

### LINKED ISSUE #1506: Typescript declaration does not include new `allowObjectInHTMLChildren` option
## 🐛 Bug Report

Following PR #1492 and issue #1483, a new option (`allowObjectInHTMLChildren`) was added to fix the `<Trans />` component. However, this seems to have been left out of the index.d.ts file, and is still raising errors inside the Trans component.

## To Reproduce

https://github.com/cravend/react-i18next-debug

## Expected behavior

The option works as expected.

## Your Environment

- *runtime version*: Typescript 4.6.3
- *i18next version*: 21.8.3, react-i18next 11.16.9
- *os*: Mac

gshokanov:
This PR fixes TypeScript error which prevents multiple nodes to be used inside core React elements when `allowObjectInHTMLChildren` is set to true.

Code example that fails on the latest published version with React 18 but works correctly with this PR:

```tsx
<Trans t={t}>
  View <b>{{ count }} cards</b>
</Trans>
```

Resolves #1506

#### Checklist

- [ ] tests are included

### DIFF
--- a/ts4.1/index.d.ts
+++ b/ts4.1/index.d.ts
@@ -102,9 +102,12 @@ declare module 'i18next' {
 type ObjectOrNever = TypeOptions['allowObjectInHTMLChildren'] extends true
   ? Record<string, unknown>
   : never;
+
+type ReactI18NextChild = React.ReactNode | ObjectOrNever;
+
 declare module 'react' {
   interface HTMLAttributes<T> {
-    children?: ReactNode | ObjectOrNever;
+    children?: ReactI18NextChild | Iterable<ReactI18NextChild>;
   }
 }
 


ts4.1/index.d.ts
```diff
@@ -100,7 +100,7 @@ declare module 'i18next' {
 }
 
 type ObjectOrNever = TypeOptions['allowObjectInHTMLChildren'] extends true
-  ? Record<string, unknown>
```
lexanth, line 103: I think this would be cleaner:
```
type HTMLChild = ReactNode | ObjectOrNever;
declare module 'react' {
  interface HTMLAttributes<T> {
    children?: HTMLChild | Iterable<HTMLChild>;
  }
}
```

I've used `Iterable` because that's what react itself uses for `ReactNode`'s circular reference, and given the type a specific name so that `ObjectOrNever` still actually represents `ObjectOrNever`.

gshokanov: True, updated MR. Thx for the suggestion!

gshokanov: Interestingly enough `Iterable` does not work correctly when combined with `t` function, but `Array` works as expected 🤔 
Here's a code example:
```tsx
<div>
    {t('hello')}
    <span>Hi</span>
</div>
```
I get a compilation error on outer div

gshokanov: Either disabling `allowObjectInHTML` or changing `Iterable` to `Array` fixes it

pedrodurek (approved):
Looks great to me! Thanks!

adrai:
included in v11.18.1

PR #1538, opened 2022-07-27
more html entities to unescape by default

adrai:
Adds more html entities to unescape without the need of a custom unescape function.
Changes also how spaces are unescaped now. Also the Trans render test needed to be adapted. Will this have an influence with existing `&nbsp;` usage?

#### Checklist

- [x] tests are included


adrai:
@cuyl can you also have a look at this?

jamuhl (approved):

PR #1539, opened 2022-07-29
Fix console warning in preact due to  Trans accessing node.children first

vivekmahajan:
- `react-i18next` when used with the latest `preact` version throws a console warning related to accessing a deprecated property (see screenshot)
-  When [upgrading from preact 8.x](https://preactjs.com/guide/v10/upgrade-guide/#notes-for-library-authors) the `VNODE` structure in `preact` was changed from `children -> props.children`
-  when `getChildren` function in `src/Trans.js` tries to access `node.children` first, `preact` throws a deprecated API console warning
- one of the ways to fix this is to follow the same pattern we do in [`hasChildren` function](https://github.com/i18next/react-i18next/blob/master/src/Trans.js#L8) i.e try to access `node.props.children` first if `node.props` exists
- This PR includes that change

image

#### Checklist
- [ ] tests are included

### DIFF
--- a/src/Trans.js
+++ b/src/Trans.js
@@ -12,7 +12,7 @@ function hasChildren(node, checkLength) {
 
 function getChildren(node) {
   if (!node) return [];
-  return node && node.children ? node.children : node.props && node.props.children;
+  return node.props ? node.props.children : node.children;
 }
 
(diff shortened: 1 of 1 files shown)


jamuhl:
thank you...will be released with next version

PR #1645, opened 2023-06-16
feat: add custom context

cqh963852:
#### Checklist

- [x] tests are included

- [ x ] motivation/reason is provided

----

While we have multiple packages. Each package has it own locale. and i18n instance.

But useTranslation can't specify which context to be use.

This solution is inspired by react-redux.


adrai:
https://github.com/i18next/react-i18next/tree/master/example/react-component-lib
https://react.i18next.com/latest/i18nextprovider#when-to-use

Issue #1601, opened 2022-12-27
Type instantiation is excessively deep and possibly infinite.ts(2589)


xdebbie:
## 🐛 Bug Report

After updating `react-i18next` from `[11.16.2]` to `[12.1.1]`, I have started getting an error on the `useTranslation` bit of the following line:

`const { t } = useTranslation('trans');`

The error is: `Type instantiation is excessively deep and possibly infinite.ts(2589)`

I can ignore it by using `/* @ts-ignore */` but that's not ideal. The error started appearing after the update and it's only on this single file (the whole app uses translation in the exact same way but strangely I am only getting this error here).

## To Reproduce

Cannot reproduce

Home.tsx
```js
import { useTranslation } from 'react-i18next';

const Home = () => {
  const { t } = useTranslation('trans');

  return (
    <div>
      <div>{t('home.firstLine')}</div>
      <div>{t('home.secondLine')}</div>
    </div>
  );
};

export default Home;
```

init-i18next.js
```js
import i18next from 'i18next';
import { initReactI18next } from 'react-i18next';
import * as fr from '@static/translations/fr.json';
import * as en from '@static/translations/en.json';

const resources = {
  fr: {
    trans: fr
  },
  en: {
    trans: en
  }
};

const lng = navigator ? navigator.language.substring(0, 2) : 'fr';

i18next.use(initReactI18next).init({
  lng, // as default
  fallbackLng: 'fr',
  defaultNS: 'trans',
  resources,
  compatibilityJSON: 'v3',
  interpolation: {
    escapeValue: false
  }
});

export default i18next;
```

i18next.d.ts
```js
import { resources, defaultNS } from './i18next';

declare module 'i18next' {
  interface CustomTypeOptions {
    defaultNS: typeof defaultNS;
    resources: typeof resources['fr'];
  }
}
```

## Expected behavior

No errors were expected, like I had before the package update.

## Your Environment

- *runtime version*: React 18.0.0
- *i18next version*: 22.4.6
- *react-i18next version*: 12.1.1
- *os*: Mac

xdebbie:
Update: I've tried all latest release versions and the problem starts occurring from `v12.0.0`

marakim:
I can confirm this problem occurs for me as well on the following configuration:
* i18next: 22.4.6
* react-i18next: 12.1.1
* react: 17.0.44
* OS: Windows

xdebbie:
[...]

PR #1646, opened 2023-06-22
Fix performance issue in `Trans` component

pedrodurek:
Closes https://github.com/i18next/react-i18next/issues/1601

Identifying the source of the problem was challenging, but thanks to a reproducible example shared by @sebastien-comeau, I was able to identify the root cause.

By default, when the `t` function is not passed, the inferred value for `KPrefix` should be `undefined`. However, due to a TypeScript limitation, the `KPrefix` was not maintaining its default signature of `undefined` and instead was inferring all keys from the `KeyPrefix`. This caused the `ParseKeys` function to map over the entire list of keys and key prefixes in order to filter out the keys that started with every `KPrefix` value. 

As a result, using the `Trans` component without specifying a `KPrefix` value could lead to an unnecessary exponential increase in the number of type inspections. This increase is due to the fact that `ParseKeys` would have to perform a large number of inspections for every possible combination of namespaces, keys, and key prefixes. Specifically, this increase in inspections can be characterized as `Namespaces * Keys * Key Prefix`. 

To avoid this performance issue, I remove the constraint value from `KPrefix` (which, in this case, brings no benefit). By removing the constraint value, `KPrefix` maintains its default value (`undefined`) and avoids unnecessary inspections.

Screenshot 2023-06-21 at 6 07 01 PM

#### Checklist

- [x] tests are included

### DIFF
--- a/TransWithoutContext.d.ts
+++ b/TransWithoutContext.d.ts
@@ -1,12 +1,4 @@
-import type {
-  i18n,
-  ParseKeys,
-  Namespace,
-  TypeOptions,
-  TOptions,
-  TFunction,
-  KeyPrefix,
-} from 'i18next';
+import type { i18n, ParseKeys, Namespace, TypeOptions, TOptions, TFunction } from 'i18next';
 import * as React from 'react';
 
 type _DefaultNamespace = TypeOptions['defaultNS'];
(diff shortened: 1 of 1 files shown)


adrai:
thank you... ❤️
included in v13.0.1

Issue #1543, opened 2022-08-15
This JSX tag's 'children' prop expects a single child of type 'ReactI18NextChild | Iterable<ReactI18NextChild>' when using `allowObjectInHTMLChildren`


davehowson:
## 🐛 Bug Report

After upgrading to React v18 and Reacti18Next to v11.18, I had an issue with objects not being accepted inside the `Trans` component, something similar to [this](https://github.com/i18next/react-i18next/issues/1504).
The fix for that was to add the `allowObjectInHTMLChildren` property, which fix that issue, but it lead to another issue.
Now, when I try to have the result of a `t` function along with some other strings inside a `p` or `span` component, I get the error

```
This JSX tag's 'children' prop expects a single child of type 'ReactI18NextChild | Iterable<ReactI18NextChild>' when using `allowObjectInHTMLChildren
````

## To Reproduce

Codesandbox - https://codesandbox.io/s/react-i18next-type-issue-uidhq7
Please check the error on `src/App.tsx:10`

```js
<p>{t("simpleContent")} with extra stuff</p>
```

## Proposed Solution

I tried playing around with the types in the library and it seems like changing `node_modules/react-i18next/ts4.1/index.d.ts:108` as follows will fix this.

```
declare module 'react' {
  interface HTMLAttributes<T> {
    children?: ReactI18NextChild | ReactI18NextChild[];
  }
}
```

The problem, at least for me, seems to be with `Iterable<ReactI18NextChild>` not working as expected for multiple children

dave-multiplier:
Hi @pedrodurek, any update on this?
I was planning on using https://github.com/ds300/patch-package to solve this for now, unless a PR is in the works?

eiskalteschatten:
We're also having this problem unfortunately. ☹️

debugelton:
same problem also with React 17 and react-i18next (v11.18.3)

JCMais:
This is probably because the type definition is `React.ReactNode | Record<string, unknown>`, while i18next type definition has `object` in the union type: https://github.com/i18next/i18next/blob/132a02461479c1b14709ff0bce1db484578ccb30/index.d.ts#L687


I think (could be wrong) that `Record<string, unknown>` and object are not compatible if you change i18next to `React.ReactNode | object` it works.

I also changed from `Iterable<ReactI18NextChild>` to `ReactI18NextChild[]`

benjamincotrel:
[...]

PR #1653, opened 2023-07-11
fix: remove override global type

marklai1998:
Fix the issue mentioned in https://github.com/i18next/react-i18next/issues/1543, react-i18next shouldn't override any global type

#### Checklist

- [x] tests are included

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -36,18 +36,6 @@ declare module 'i18next' {
   }
 }
 
-type ObjectOrNever = TypeOptions['allowObjectInHTMLChildren'] extends true
-  ? Record<string, unknown>
-  : never;
-
-type ReactI18NextChildren = React.ReactNode | ObjectOrNever;
-
-declare module 'react' {
-  interface HTMLAttributes<T> {
-    children?: ReactI18NextChildren;
-  }
-}
-
 type _DefaultNamespace = TypeOptions['defaultNS'];
 
 export function useSSR(initialI18nStore: Resource, initialLanguage: string): void;


pedrodurek:
Hey @marklai1998, I'd like to clarify a couple of things.

1. We're not overriding any type, and instead utilizing the powerful feature of **TypeScript** known as "type augmentation" to **extend** an interface. Through this, we're "extending" further capabilities to the `HTMLAttributes` interface from the **React** library. Since React 18, changes were made leading the type definitions to stop accepting objects as children.
   
2. We're not overriding any global types. To be precise, it's the `HTMLAttributes` interface from the **React** library that's being **extended**. Although global types' extension isn't necessarily a problematic practice, it's notable that it's commonly embraced by numerous libraries, including React itself.
   
3. Setting `allowObjectInHTMLChildren` as `false` won't extend the original `HTMLAttributes` interface type. The extension remains limited to the `children` type alone. The original type for `HTMLAttributes` doesn't change.

In case we opt-out of type augmentation, we'll need to notify the TypeScript users that implementing objects (`{{name}}`) as children in the `Trans` component is no longer an acceptable practice. 

Lastly, could you provide any information where it says that a library shouldn't **extend** types from another library or global types? The main purpose of type augmentation is to adapt types from third-party libraries when needed.

marklai1998:
@pedrodurek thanks for your comment, maybe I'm not that familiar with typescript technical terms
But this file clearly override the react HTMLAttributes children property

whenever react-i18next is imported, the .d.ts will kicks in and "override"(alter the behavior) the type globally
where you cannot really turn it off via any js code
`declare module` exist = modify the global

Let me spend some time on a minimal reproduction of our issue in the issue page

Meanwhile, can you also create an example of what this trying to solve?
I don't see any type error under `test/typescript/*` after I removed the type
I'm sure we can find the best way to solve this issue, thanks

PR #1660, opened 2023-08-08
Fix non-list dynamic content in Trans component

### LINKED ISSUE #1658: Allow non-list children to render correctly with i18nIsDynamicList
## 🚀 Feature Proposal / Possible bug fix
Hi there!

Proposals (all of which I can open a PRs for, if desired):
1. Allow `i18nIsDynamicList` to function correctly, even when child is not a list.
2. Rename `i18nIsDynamicList` to `i18nIsDynamicContent` (will be done in a backwards compatible way, i.e. both will exist for a time, maybe till v14)

## Motivation
1. I believe this first proposal is actually a bugfix. Currently, the Trans component assumes that the child of an element with the `i18nIsDynamicList` prop is a list, which may not be the case. See the following example:
```ts
const someDynamicVar = cond ? <Comp /> : 'Text';

<Trans>
  Hi there,{' '}
  <React.Fragment i18nIsDynamicList>
    {someDynamicVar}
  </React.Fragment>
</Trans>
```
The above will not render correctly. However, the following does:
```ts
const someDynamicVar = cond ? <Comp /> : 'Text';

<Trans>
  Hi there,{' '}
  <React.Fragment i18nIsDynamicList>
    <></>
    {someDynamicVar}
  </React.Fragment>
</Trans>
```

By adding a fragment, the child becomes a list, and therefore the content renders correctly. My proposal is to make it work without the need for a list.

2. If we make the above change, which seems to me to be a bugfix, it would make sense to rename the `i18nIsDynamicList` prop to something that better describes what it's for. I was thinking `i18nHasDynamicChildren`. We can make this change in a backwards compatible way (i.e. both `i18nIsDynamicList` and `i18nHasDynamicChildren` will exist for a time, maybe till v14)

Please let me know if these changes are welcome, in which case I'll make a PR!

nicegamer7:
Hi! This PR closes #1658.

#### Changes
- Always return children as an array from getChildren (for isI18nDynamicList)
  - This is the bugfix
- Reduce false positives when checking for nodes needing handling with empty children
  - This is just something I noticed while working
- Remove React warnings in Trans component by removing i18nIsDynamicList prop while rendering
  - Again, this is just something I noticed and thought to fix while I was at it

#### Checklist

- [x] run tests `yarn test`
- [x] tests are included

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -36,6 +36,14 @@ declare module 'i18next' {
   }
 }
 
+declare global {
+  namespace JSX {
+    interface IntrinsicAttributes {
+      i18nIsDynamicList?: boolean;
+    }
+  }
+}
+
 type ObjectOrNever = TypeOptions['allowObjectInHTMLChildren'] extends true
   ? Record<string, unknown>
   : never;
--- a/src/TransWithoutContext.js
+++ b/src/TransWithoutContext.js
@@ -138,7 +138,6 @@ function renderNodes(children, targetString, i18n, i18nOptions, combinedTOpts, s
   function renderInner(child, node, rootReactNode) {
     const childs = getChildren(child);
     const mappedChildren = mapAST(childs, node.children, rootReactNode);
-    // console.warn('INNER', node.name, node, child, childs, node.children, mappedChildren);
     return hasValidReactChildren(childs) && mappedChildren.length === 0 ? childs : mappedChildren;
   }
 
@@ -149,8 +148,9 @@ function renderNodes(children, targetString, i18n, i18nOptions, combinedTOpts, s
     } else {
       mem.push(
         ...Children.map([child], (c) => {
-          const { i18nIsDynamicList, ...rest } = c.props;
-          return <c.type key={i} ref={c.ref} {...rest} {...(isVoid ? {} : { children: inner })} />;
+          const props = { ...c.props };
+          delete props.i18nIsDynamicList;
+          return <c.type {...props} key={i} ref={c.ref} {...(isVoid ? {} : { children: inner })} />;
         }),
       );
     }
@@ -171,11 +171,15 @@ function renderNodes(children, targetString, i18n, i18nOptions, combinedTOpts, s
         i18n.services.interpolator.interpolate(node.children[0].content, opts, i18n.language);
 
       if (node.type === 'tag') {
-        let tmp = reactNodes[parseInt(node.name, 10)]; // regular array (components or children)
-        if (!tmp && rootReactNode.length === 1 && rootReactNode[0][node.name])
-          tmp = rootReactNode[0][node.name]; // trans components is an object
-        if (!tmp) tmp = {};
-        //  console.warn('TMP', node.name, parseInt(node.name, 10), tmp, reactNodes);
+        // regular array (components or children)
+        let tmp = reactNodes[parseInt(node.name, 10)];
+
+        // trans components is an object
+        if (rootReactNode.length === 1) tmp ||= rootReactNode[0][node.name];
+
+        // neither
+        tmp ||= {};
+
         const child =
           Object.keys(node.attrs).length !== 0 ? mergeProps({ props: node.attrs }, tmp) : tmp;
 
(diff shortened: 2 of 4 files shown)


src/TransWithoutContext.js
```diff
@@ -1,4 +1,4 @@
-import { isValidElement, cloneElement, createElement } from 'react';
+import React, { isValidElement, cloneElement, createElement, Children } from 'react';
```
adrai, line 1: Why is the React import necessary?

index.d.ts
```diff
@@ -36,6 +36,14 @@ declare module 'i18next' {
   }
 }
 
+declare global {
+  namespace JSX {
+    interface IntrinsicAttributes {
+      i18nIsDynamicList?: boolean;
```
adrai, line 42: Why is this TS change necessary?

src/TransWithoutContext.js
nicegamer7: The React import allows us to use JSX in the file (used inside the Children.map)

index.d.ts
nicegamer7: It allows you to pass the i18nIsDynamicList to the React.Fragment component.

src/TransWithoutContext.js
adrai: ahh, yes

jamuhl:
What is the use case of having `i18nIsDynamicList` in combination with a non-list?

You give the sample:


```js
<Trans>
  Hi there,{' '}
  <React.Fragment i18nIsDynamicList>
    {someDynamicVar}
  </React.Fragment>
</Trans>
```

which has nothing todo with the usage of `i18nIsDynamicList`

that flag is only used for creating the string for saveMissing feature and only for the case where content is like: https://react.i18next.com/latest/trans-component#using-with-lists-v10.5.0

nicegamer7:
In our project we're using key fallback for all translations, so the key for `<Trans>Hi</Trans>` would be `"Hi"`. This worked well for 99% of our project, but in a couple places, we had dynamic variables being used, and these weren't being extracted correctly by `i18next-parser` (rather than extracting as `<3></3>` for example, it extracted as `{someDynamicVar}`). This caused the translations to not get matched properly to their respective Trans elements, so I thought this was the expected behaviour since it worked with fine with `i18nIsDynamicList` (after this bug fix, and another to `i18next-parser`).

It seems extracting JSX expressions is broken in `i18next-parser`, so I'll get that fixed in another PR. Regardless though, there is still a valid use for this PR which I came across in another file yesterday:
```
<Trans>
  <React.Fragment isI18nDynamicList>
    {list.length === 1 ? list[0] : list.map((x) => <div> ... </div>)}
  </React.Fragment>
</Trans>
```

We have a couple different pieces of code that are similar to the above in our project, where we want to display the list differently depending on whether it has 1 element, or multiple. We could split it into a ternary with two Trans elements, but this reduces duplicate code.

src/TransWithoutContext.js
```diff
@@ -13,7 +13,7 @@ function hasChildren(node, checkLength) {
 
 function getChildren(node) {
   if (!node) return [];
-  return node.props ? node.props.children : node.children;
+  return getAsArray(node.props ? node.props.children : node.children);
```
adrai, line 16: I would not always return an array... but maybe only if i18nIsDynamicList

jamuhl: Or assert you always pass an array when using `i18nIsDynamicList` (as the name suggests it's about lists and not single elements) -> code has to look like:

```js
   <Trans>
      <React.Fragment i18nIsDynamicList>
        {list.length === 1 ? [<div>{list[0]}</div>] : list.map((x) => <div> {x} </div>)}
      </React.Fragment>
    </Trans>
```

nicegamer7: Thinking about this more, the following would produce the exact same result:
```ts
    <Trans>
      {list.length === 1 ? list[0] : list.map((x) => <div>{x}</div>)}
    </Trans>
```

Am I correct in the above?

Either way, I think allowing single elements would make this prop easier to use. I don't quite see any reasoning why disallowing single elements would make sense, other than the fact that the prop mentions the word "list" (which I think would be better named `i18nHasDynamicContent` or something similar).

Let me know the final decision though, and I can change/close this PR accordingly.

adrai: for sure I would not always return an array, but if, then only for i18nIsDynamicList

adrai:
@nicegamer7 shouldn't that warning be gone?
image

adrai:
btw: included in v13.1.0

nicegamer7:
@adrai I wasn't able to remove the warning from React.Fragment, since it has its own validation that happens before rendering occurs. But the warning is removed for all HTML elements.

PR #1661, opened 2023-08-10
Render all children regardless of type when using i18nIsDynamicList prop

nicegamer7:
Hi @adrai, this is followup to #1660. I noticed that rendering was still not working when using dynamic strings. This PR fixes that.

#### Checklist
- [x] tests are included

### DIFF
--- a/src/TransWithoutContext.js
+++ b/src/TransWithoutContext.js
@@ -139,7 +139,12 @@ function renderNodes(children, targetString, i18n, i18nOptions, combinedTOpts, s
   function renderInner(child, node, rootReactNode) {
     const childs = getChildren(child);
     const mappedChildren = mapAST(childs, node.children, rootReactNode);
-    return hasValidReactChildren(childs) && mappedChildren.length === 0 ? childs : mappedChildren;
+    // `mappedChildren` will always be empty if using the `i18nIsDynamicList` prop,
+    // but the children might not necessarily be react components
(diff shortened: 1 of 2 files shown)


adrai:
v13.1.1

siarheipashkevich:
After upgrading to the latest version:
image

adrai:
@siarheipashkevich Please provide a reproducible example repo.

siarheipashkevich:
I just upgraded my packages and try to start my build in web

adrai:
Sorry, I'm not able to reproduce it here: https://github.com/i18next/react-i18next/tree/master/example/react

siarheipashkevich:
I use react-native and expo

nicegamer7:
@siarheipashkevich it looks like your browser is not expecting the [optional chaining operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining). If this only happens in development, I'd suggest making sure your browser is up to date. If this happens both in development and in production, I would suggest transpiling your code using something like Babel.

siarheipashkevich:
After downgrading to `13.0.3` everything works correctly

siarheipashkevich:
I think you don't use `optional chaining` operator for supporting `ES5`

adrai:
Does it work with  13.1.2 ?

siarheipashkevich:
Yes, it's work correctly right now, thanks!

PR #1664, opened 2023-08-22
Don't use `defaults` prop as default key

### LINKED ISSUE #1663: Unable to provide default value for Trans without i18nKey
## 🐛 Bug Report
At the company I'm working at, we're using key fallback for all of our translations as this allows us to keep the strings inside the code. We're also using the pluralize library to allow us to leave our English resource file empty.

Here's an example using `t`:
```ts
t('You have {{count}} message', {
  count,
  defaultValue: `You have ${count} ${pluralize('message', count)}`,
});
```

The above works fine with `t`, but does not work with the Trans component since the key defaults to the `defaults` prop.

## To Reproduce
```ts
<Trans
  count={count}
  defaults={`You have ${count} ${pluralize('message', count)}`}
>
  You have {{ count }} message
</Trans>
```

## Expected behavior
I expect the key to be `You have {{count}} message`, but the key that react-i18next uses in the `t` call is `You have 3 messages` (if `count` is 3, for example).

## Proposed solutions
1. **(breaking change)** Rather than defaulting the key to the `defaults` prop [here](https://github.com/i18next/react-i18next/blob/52d514c64ac15dadfe5cb263918c0d5ce06125e4/src/TransWithoutContext.js#L337), default to `nodesToString(children, reactI18nextOptions)`.
2. Add a `defaultValue` prop that behaves similar to the one in `t` (i.e. does not get used as the key).

I can open a PR for this if we decide on a solution.

nicegamer7:
Fixes #1663.

#### Checklist
- [x] tests are included

### DIFF
--- a/src/TransWithoutContext.js
+++ b/src/TransWithoutContext.js
@@ -328,13 +328,11 @@ export function Trans({
   let namespaces = ns || t.ns || (i18n.options && i18n.options.defaultNS);
   namespaces = typeof namespaces === 'string' ? [namespaces] : namespaces || ['translation'];
 
+  const nodeAsString = nodesToString(children, reactI18nextOptions);
   const defaultValue =
-    defaults ||
-    nodesToString(children, reactI18nextOptions) ||
-    reactI18nextOptions.transEmptyNodeValue ||
-    i18nKey;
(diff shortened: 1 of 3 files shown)


adrai:
thank you...
should be included in v13.2.0

PR #1669, opened 2023-08-29
Allow iterable `ReactI18NextChildren` as children

nicegamer7:
Fixes issues where interpolations alongside other HTML element children would trigger TypeScript errors. Added tests to make sure this stays working.

Note: I went with React 18 typings becuase React 16 typings allowed for objects to be passed as children, which made it hard to test that our `allowObjectInHTMLChildren` option was working correctly. This shouldn't affect end-users since it's a development dependency.

#### Checklist
- [x] tests are included

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -52,7 +52,10 @@ type ReactI18NextChildren = React.ReactNode | ObjectOrNever;
 
 declare module 'react' {
   interface HTMLAttributes<T> {
-    children?: ReactI18NextChildren;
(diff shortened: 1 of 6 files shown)


adrai:
thank you @nicegamer7
v13.2.1

nicegamer7:
No worries :)

YutaMoriJP:
@adrai @nicegamer7 Thanks you two 🙏

PR #1684, opened 2023-10-12
fix(context): add pure annotation to allow tree-shaking

Sarabadu:
Apparently the createContext prevents the file from been tree shaked from the esbuild bundler 

i created a playground [here](https://hyrious.me/esbuild-repl/?version=0.19.4&b=e%00entry.js%00import+%7BB%7D+from+%27.%2Ffile2.js%27%0A%0Aconsole.log%28B%28%29%29&b=%00file.js%00import+%7BcreateContext%7D+from+%27react%27%0A%0Aexport+const+context+%3D+%2F*+%23__PURE__+*%2F+createContext%28%29%3B%0A%0Aexport+const+A+%3D+%27a%27%3B&b=%00file2.js%00import+%7BA%7D+from+%27.%2Ffile.js%27%0A%0Aexport+const+B+%3D+function+%28%29%7B%0A++return+%27hello+world%27%0A%7D&i=react%4018.2.0&o=--platform%3Dbrowser+--format%3Desm+--bundle+--external%3Areact+) to replicate my issue.

when removing the `/* #__PURE__ */` annotation it doesn't matter that the context is not used anywhere, the bundle cannot treeShake.

I saw similar similar changes on other libraries to improve tree-shaking ([ref](https://github.com/reduxjs/react-redux/issues/1470))

#### Checklist

- [ ] tests are included

### DIFF
--- a/src/context.js
+++ b/src/context.js
@@ -3,9 +3,9 @@ import { getDefaults, setDefaults } from './defaults.js';
 import { getI18n, setI18n } from './i18nInstance.js';
 import { initReactI18next } from './initReactI18next.js';
 
-export { getDefaults, setDefaults, getI18n, setI18n, initReactI18next };
+export { getDefaults, getI18n, initReactI18next, setDefaults, setI18n };
 
-export const I18nContext = createContext();
+export const I18nContext = /* #__PURE__ */ createContext();
 
 export class ReportNamespaces {
   constructor() {


adrai:
Can you explain exactly why this is necessary? Maybe a link that explains this more in detail?

Sarabadu:
yes sorry for the poor description. analysing my bundled code I realised that this context creation is added to the bundle even if there is no usage, of it.
that is because the createContext add a side effect, i will make this a draft because I still need to corroborate if there is no other way to remove this

Sarabadu:
@adrai added a bit of context for it, hope it makes sense

adrai:
Are you using the Trans component or are you using the TransWithoutContext?
Because normally if you use the Trans component or the useTranslation hook it should be bundled anyway...
Only on server side working without context may make sense.

Sarabadu:
We are not using this library on our bundle, (bundling for a service worker file), we remove all the react code from there but this is detected as a sideEffect and cannot be removed

jamuhl:
why you use react-i18next on a worker? that makes no sense to me...

translation should occure close to the rendering - not inside some sandboxed worker

Sarabadu:
We are not using it at all, but get bundles anyway because of this

adrai:
Can you create a minimal reproducible example of your setup so we may understand better?

PR #1707, opened 2024-01-02
test: migrate ts test to vitest

marcalexiei:
Followup of i18next/i18next#2097.

Now all typescript test run with `vitest` using the same approach of i18next.

#### Checklist

- [x] only relevant code is changed (make a diff before you submit the PR)
- [x] run tests `npm run test`
- [x] tests are included
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/i18next/.github/blob/master/CONTRIBUTING.md)

PR #1708, opened 2024-01-02
test: migrate to vitest

marcalexiei:
Followup of #1707.

All tests now run with `vitest`. 

~~`TransWithoutContext ` file have been renamed with `jsx` extension otherwise `vitest` is not able to compile it correctly. 
Regular build should not be affect by this change.~~

[`TransWithoutContext` now use React functions to create JSX.Element ](https://github.com/i18next/react-i18next/pull/1708#discussion_r1439355813)

#### Checklist

- [x] tests are included


src/index.js
```diff
@@ -1,5 +1,5 @@
 export { Trans } from './Trans.js';
-export { Trans as TransWithoutContext } from './TransWithoutContext.js';
+export { Trans as TransWithoutContext } from './TransWithoutContext.jsx';
```
adrai, line 2: can we keep this js instead of jsx?

marcalexiei: > TransWithoutContext file have been renamed with jsx extension otherwise vitest is not able to compile it correctly.

`vitest` complains when compiling a js file which contains actual jsx (rightfully IMHO 😁)

image

----

> Regular build should not be affect by this change.

I have executed `npm run build` and dist files are not modified so I don't think that this change will affect library usage

adrai: I don't think the dist files are not modified... for sure the path to .tsx is changed...
I like to keep .js if possible

marcalexiei:
I'm confused by code climate check:
all the reported issues were already present in the file so why the check fails now? 😑

Do I need to implement the provided suggestions?

adrai:
> I'm confused by code climate check: all the reported issues were already present in the file so why the check fails now? 😑
> 
> Do I need to implement the provided suggestions?

no, these issues are there because of the filename change... you can ignore them

src/index.js
adrai: maybe we can remove the jsx parts:

image
image

adrai:
```js
import { isValidElement, cloneElement, createElement, Children, Fragment } from 'react';
// ...
          return createElement(c.type, { ...props, key: i, ref: c.ref, ...(isVoid ? {} : { children: inner }) });
          // return <c.type {...props} key={i} ref={c.ref} {...(isVoid ? {} : { children: inner })} />;
// ...
      function Componentized() {
        return createElement(Fragment, null, comp);
        // return <>{comp}</>;
      }
      components[c] = createElement(Componentized, null);
      // components[c] = <Componentized />;
```

marcalexiei: > I don't think the dist files are not modified... for sure the path to .jsx is changed...
> I like to keep .js if possible

`dist` files have always `js` extension because they are compiled by `rollup`. 
You can't have jsx file inside dist. (Try run the build on `master` and this PR branch)

> maybe we can remove the jsx parts:

Sure, I can make this change.

adrai:
Do you see the same errors locally?

marcalexiei:
> Do you see the same errors locally?

No, on my machine works fine: 

image

adrai:
I suspect there's a caching issue on circleci... can you update the package.json i.e. by updating the i18next dependency?
image

adrai:
looks good now

marcalexiei:
Yeah! Nice catch!

adrai:
strange it fails again: https://app.circleci.com/pipelines/github/i18next/react-i18next/1193/workflows/df08dd70-3de3-44b9-ac09-76d291c28777/jobs/1066

adrai:
ok, the coverage part seems to be the problem, like described here: https://github.com/vitest-dev/vitest/issues/317

marcalexiei:
I’ll take a look later (having dinner)

adrai:
fixed with: https://github.com/i18next/react-i18next/commit/21e773884f9903fc23abbbde11aae363f0e46012#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R150
https://app.circleci.com/pipelines/github/i18next/react-i18next/1195/workflows/ff1fa60f-8aa7-42e8-825e-c8b9222c8db4/jobs/1070

marcalexiei:
Great! Thank you for looking into this 🎉

Issue #1691, opened 2023-10-31
t function from useTranslation is unstable in concurrent mode


woodreamz:
## 🐛 Bug Report

The bug is really well explained in #1599. This issue has been also reported in #1636. Those issues have been closed without being fixed.

When enabling concurrent mode in React, the `t` function changes and creates extra renders. i18n.t is not doing that. For now, it's not major as concurrent mode is not enabled in production build but could become major in a futur react release.

The issue seems to be in `useTranslation`. I am trying to find a fix and propose a PR. This issue will help to track the work.

For now, concurrent mode is enabled with [StrictMode](https://react.dev/reference/react/StrictMode). 

> You can use [<StrictMode>](https://react.dev/reference/react/StrictMode) to help surface concurrency-related bugs during development. Strict Mode doesn’t affect production behavior, but during development it will log extra warnings and double-invoke functions that are expected to be idempotent.
Source: [react.dev](https://react.dev/blog/2022/03/29/react-v18#gradually-adopting-concurrent-features)

## To Reproduce

In this [codesandbox](https://codesandbox.io/s/react-i18next-concurrent-mode-wjrtjt), you can se in the console that the useEffect on i18n is called **2** times and the useEffect on t is called **3** times.

```js
import "./i18next";
import { useTranslation } from "react-i18next";
import { useEffect } from "react";

export default function App() {
  const { i18n, t } = useTranslation();

  useEffect(() => {
    console.log("useEffect with i18n");
  }, [i18n]);

  useEffect(() => {
    console.log("useEffect with t");
  }, [t]);

  return i18n.t("text");
}
```

## Expected behavior

useEffect with t as dependency should be called **2** times.

woodreamz:
It seems to be caused by: https://github.com/i18next/react-i18next/blob/461986289fe82a70a5ddae9078c7c7e68f86f7dd/src/useTranslation.js#L117

qq831799:
Same issue here, and my react-i18next version is 12.3.1

timheilman:
same issue here. Thank you @woodreamz for your sleuthing so far.

adrai:
> It seems to be caused by:
> 
> https://github.com/i18next/react-i18next/blob/461986289fe82a70a5ddae9078c7c7e68f86f7dd/src/useTranslation.js#L117

[...]

PR #1716, opened 2024-02-01
Fix/bug 1691 make returned t function identical upon second effect run in strict mode

timheilman:
#### Checklist

- [ ] tests are included


adrai:
@medihack can you review this?

timheilman:
https://github.com/testing-library/react-testing-library/pull/1241 was _just_ merged, but I haven't looked closely yet at it to see if hook-based testing support was added for useEffect double-run in strictMode

timheilman:
OK, I went down a bit of a rabbit hole toward writing a unit test for this, but it looks like strictMode is integrated into @testing-library/react (which subsumed @testing-library/react-hooks) only for v18 of react and higher, whereas react-i18next is still supporting v16 of react. That precludes using the strictMode testing configuration, without which it's hard for me to see how to test this change otherwise than as described in the "To Reproduce" section of the [issue](https://github.com/i18next/react-i18next/issues/1691). That reproduction is reliable and is indeed fixed by this change.

adrai:
ok, no test required, but I like @medihack to check if this change will influence in some way the issue he had [here](https://github.com/i18next/react-i18next/issues/1272#issuecomment-791745096)

adrai:
//cc: @kachkaev

timheilman:
> ok, no test required, but I like @medihack to check if this change will influence in some way the issue he had [here](https://github.com/i18next/react-i18next/issues/1272#issuecomment-791745096)

I totally understand.  This is a widely-depended-upon library that I'm certainly no expert with. I welcome any/all code review comments and/or edits by those more familiar with the code.

The test @medihack added alongside the change for the issue of changing the provider-provided i18n instance _is_ still passing, which is a good sign. This makes sense since the `useCallback` memoization _does_ have that `i18n` instance in its dependency list, and so should generate a new fixedT. However, the truth is that I'm not very familiar at all with this code, and it is hairy, so I still very much welcome more experienced eyes on it.

medihack:
At first glance, it looks good. If there is no time pressure, I would take a deeper look and try it out on the weekend if that is ok for you.

adrai:
wonderful... thank you @medihack

medihack:
The PR looks good to me. I also gave it a quick test drive and got no problems (with shallow routing in Next.js).

adrai:
Thank you very much for your contributions... it's included in v14.0.2

jaredhan418:
Why!!!!!! Why!!!!! Damn!
`i18n.changeLanguage` will never work immediately in every env!
StrictMode is so damn bad!

adrai:
@jaredhan418 is right... this PR breaks the language change rendering... @timheilman can you have a look at it? @medihack can you assist?

adrai:
I temporarily reverted it with v14.0.3

medihack:
That's unfortunate :-( And unfortunately, I don't have enough time right now. I'll try my best in the upcoming weeks, but it looks like a nasty issue to fix entirely.

timheilman:
happy to take a look! Though of course I can't promise anything.  Thanks @jaredhan418 for your testing and notifying of the breakage, and thanks @adrai for being so on-top of this repo.  Reversion is obviously the right move.

Issue #1718, opened 2024-02-05
Changing the language doesn't change the text in react native


gsk143:
## 🐛 Bug Report

Changing the language doesn't change the text in react native

## To Reproduce
I have created a bare RN app and tried following the steps in example folder. For ease I have added an attachment
## Expected behavior
Change in language should change the text 

## Your Environment

- *runtime version*: i.e. node v20.11.0
- *i18next version*: i.e. 23.8.2
- *os*: Mac
- *Android Emulator API  - 33*
[LanguageChange.zip](https://github.com/i18next/react-i18next/files/14161872/LanguageChange.zip)

I am new to RN so any help is appreciated

jaredhan418:
Not your fault. This [PR](https://github.com/i18next/react-i18next/pull/1716) brings this issue.

adrai:
yes, try v14.0.1

gsk143:
Yep that fixed it. Thanks

adrai:
reverted with v14.0.3 and waiting for a better fix

adrai:
please try also v14.0.5

PR #1720, opened 2024-02-06
Fix/bug 1691 preserving change language binding

timheilman:
#### Checklist

- [x] tests are included


timheilman:
Sorry about the initial bad fix in #1716 .  A dev issue was uncovered over at https://github.com/i18next/react-i18next/pull/1716#issuecomment-1926458005 . Unfortunately I still cannot figure out how to test the bug 1691 proper in a unit test, but the dev issue does have a test added that:

1) succeeds before #1716 
2) fails after #1716 , demonstrating the dev issue
3) succeeds after this PR

I also have tested this PR manually for both bug 1691 proper as well as for the dev issue. I welcome @adrai and @medihack 's eyes on this second attempt, as well as anyone else's interested.  No time pressure.

adrai:
thank you... it's included in v14.0.5

adrai:
fyi: verified to work also in this example https://github.com/i18next/react-i18next/commit/8334298fa563ab913ced244b26438285067a8cbc 👍

timheilman:
If `14.0.1` exhibits the bug #1756 , then the fix to #1756 in `14.0.2` comes at the cost of causing bug #1718 , which is why the change in `14.0.2` was rolled-back in `14.0.3`.  Please see [my recent comment on issue 1756](https://github.com/i18next/react-i18next/issues/1756#issuecomment-2156706232).

If that is the case, then I would ask @schontz to provide a minimal reproduction with expected and actual results, similar to the way that @woodreamz did in #1691 .  That "To Reproduce" section with its behavior that reliably differs from the "Expected behavior" is the gold standard in bug reports.  If you can file an issue that way, I've been very impressed with @adrai 's responsiveness to PRs provided by the community, and your concern has a good chance to be addressed.

PR #1731, opened 2024-03-06
fix: issue 1728 when useSuspense is false and default ns

### LINKED ISSUE #1728: 14.0.5 with custom `defaultNS` and ` useSuspense: false` does not work with `i18next-resources-to-backend`
## 💥 Regression Report

Keys are not loading on refresh anymore, instead the raw key name is displayed. It seems to be an edge case with our combination of settings (see title and repro-steps).

This applies to the dev environment with vite, I haven't tested it in any production setting.

One curious thing is that when doing any change to the template to trigger a hot module reload, the key starts showing.

## Last working version

Worked up to version: 14.0.4

Stopped working in version: 14.0.5

## To Reproduce

I've created a repository with the minimal possible setup that shows the issue. It's based on vite / react-ts. I've included a commit that reverts to 14.0.4 where the setup is working fine.

https://github.com/arechsteiner/react-i18next-repro

The main parts are:
- using `resouresToBackend`
- `useSuspense: false` <- removing this will make it work
- `defaultNS: 'common'` <- removing this will also make it work

```js
import i18n from 'i18next'
import { useTranslation, initReactI18next } from 'react-i18next'
import resourcesToBackend from 'i18next-resources-to-backend'
import en from './locales/en/translation.json'

i18n
  .use(
    resourcesToBackend((language, namespace, callback) => {
      console.log('namespace', namespace, language)
      callback(null, en)
    }),
  )
  .use(initReactI18next)
  .init({
    defaultNS: 'common',
    lng: 'en',
    fallbackLng: 'en',
    react: {
      useSuspense: false,
    },
    interpolation: {
      escapeValue: false,
    },
  })
```

## Expected behavior

The key should show it's translation right away as in 14.0.4

## Your Environment

- *runtime version*: Node v18.14.2, Chrome 122.0.6261.94
- *i18next version*: ^23.10.0
- *os*: Mac

timheilman:
is custom, in strict mode only, memoization is performed of an unready translation function and is not refreshed after readiness.

The use of the isInitial ref is fundamentally incompatible with strict mode. Substituting ready instead seems to fix #1728 and keep both #1691 and #1718 fixed

#### Checklist

- [ ] tests are included


timheilman:
A review from @medihack , especially for the shallow routing issue he was familiar with. It's a shame it's so difficult to unit-test StrictMode.

PR #1761, opened 2024-07-15
use the commented out async code

VIKTORVAV99:
There was some code that was commented out that should be more performant and smaller than the existing code. This was done to avoid pulling in regenerator runtimes in the past but it seems this is no longer an issue and can be used without doing so.

#### Checklist

- [x] tests are included

### DIFF
--- a/src/context.js
+++ b/src/context.js
@@ -24,35 +24,18 @@ export class ReportNamespaces {
 }
 
 export function composeInitialProps(ForComponent) {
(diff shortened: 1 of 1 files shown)


adrai:
Do you need a new release with this or can this wait?

VIKTORVAV99:
> Do you need a new release with this or can this wait?

This can wait, I have some other PRs planned as well 🙂

PR #1764, opened 2024-07-15
Simplifies hasValidReactChildren

VIKTORVAV99:
Simplifies `hasValidReactChildren` a bit by utilizing Array.isArray, implicit returns and the fact that `.every()` can take a function without arguments as it's own argument.

This no longer returns false when it's not an array but rather implicitly returns `undefined` which don't affect current functionality as the code it's used in just checks for truthy values.

#### Checklist

- [x] tests are included

### DIFF
--- a/src/TransWithoutContext.js
+++ b/src/TransWithoutContext.js
@@ -17,10 +17,8 @@ const getChildren = (node) => {
   return node.props && node.props.i18nIsDynamicList ? getAsArray(children) : children;
 };
 
-const hasValidReactChildren = (children) => {
-  if (Object.prototype.toString.call(children) !== '[object Array]') return false;
-  return children.every((child) => isValidElement(child));
-};
+const hasValidReactChildren = (children) =>
+  Array.isArray(children) && children.every(isValidElement);
 
 const getAsArray = (data) => (Array.isArray(data) ? data : [data]);
 


VIKTORVAV99:
@adrai how do you feel about making some changes that might potentially be breaking changes?

I'm thinking about optional chaining and removing the code here:
https://github.com/i18next/react-i18next/blob/92bc14704e447ac38973b804e852c7ccd6c1c15a/src/utils.js#L57-L96


Both of these changes should be supported by the current build config as far as I can tell. Browserslist is configured to only target the `default` browsers and the code in question is never used if the peer dependency version of i18next is followed.

adrai:
same topic like https://github.com/i18next/i18next/pull/2184

VIKTORVAV99:
What about removing the old code then? That has nothing to do with system compatibility and should be fine as long as the peer versions are followed.

adrai:
yes, I think 3 years later, we can remove that old code for a future major version (also conditional chaining)...
as soon as you "finished" to create other PRs... (non-breaking) I will create a NON-major version, and afterwards we can go for a new major version, ok?

VIKTORVAV99:
Sounds good!

I'll go through the code a final time and see if I find any other small opportunities to optimise the current code without breaking changes.

PR #1782, opened 2024-08-12
chore: update test deps

ripecosta:
### Description

Update all test related dependencies possible at this stage. See comments for more info.

#### Checklist

- [x] tests are included


package.json
```diff
@@ -84,15 +84,16 @@
     "@rollup/plugin-node-resolve": "^15.2.3",
     "@rollup/plugin-replace": "^5.0.7",
     "@rollup/plugin-terser": "0.4.4",
-    "@testing-library/jest-dom": "^5.16.5",
-    "@testing-library/react": "^11.2.7",
-    "@testing-library/react-hooks": "^3.4.2",
-    "@types/react": "^18.2.52",
-    "@vitest/coverage-v8": "^1.2.2",
+    "@testing-library/jest-dom": "^6.4.8",
+    "@testing-library/react": "^12.1.5",
```
ripecosta, line 88: @testing-library/react v13 drop support for react v17 and below so can't upgrade further unless this project also projects support for react v16.

package.json
```diff
@@ -84,15 +84,16 @@
...
+    "@testing-library/react-hooks": "^8.0.1",
+    "@types/jest": "^29.5.12",
+    "@types/react": "^16.14.60",
+    "@vitest/coverage-v8": "^2.0.5",
     "all-contributors-cli": "^6.26.1",
     "babel-core": "^7.0.0-bridge.0",
-    "babel-plugin-macros": "^2.8.0",
-    "babel-plugin-tester": "^6.5.0",
+    "babel-plugin-macros": "^3.1.0",
+    "babel-plugin-tester": "^7.0.4",
```
ripecosta, line 96: Updating babel-plugin-preset produces a lot of snapshot changes because of how it formats them so it's best done in a separate PR.

package.json
```diff
@@ -84,15 +84,16 @@
...
     "@rollup/plugin-terser": "0.4.4",
-    "@testing-library/jest-dom": "^5.16.5",
-    "@testing-library/react": "^11.2.7",
-    "@testing-library/react-hooks": "^3.4.2",
-    "@types/react": "^18.2.52",
-    "@vitest/coverage-v8": "^1.2.2",
+    "@testing-library/jest-dom": "^6.4.8",
+    "@testing-library/react": "^12.1.5",
+    "@testing-library/react-hooks": "^8.0.1",
+    "@types/jest": "^29.5.12",
```
ripecosta, line 90: @testing-library/jest-dom v6 no longer has @types/jest as its dependency so we need to import it explicitly now.

test/setup.js
```diff
@@ -1,4 +1 @@
-import matchers from '@testing-library/jest-dom/matchers';
-import { expect } from 'vitest';
-
-expect.extend(matchers);
+import '@testing-library/jest-dom/vitest';
```
ripecosta, line 1: See https://github.com/testing-library/jest-dom?tab=readme-ov-file#with-vitest

test/typescript/misc/Trans.test.tsx
```diff
@@ -126,7 +126,6 @@ describe('<Trans />', () => {
   it('should not work with object child', () => {
     assertType<React.ReactElement>(
       <Trans>
-        {/* @ts-expect-error */}
```
ripecosta, line 129: No longer necessary after downgrading @types/react to v16.

package.json
```diff
@@ -84,15 +84,16 @@
...
+    "@types/react": "^16.14.60",
```
ripecosta, line 91: Downgrade @types/react to v16 was necessary because of @testing-library/react-hooks v8. Regardless, it makes sense to use v16 instead of v18 because that's what this repo uses to test.

package.json
```diff
@@ -116,7 +117,7 @@
     "rimraf": "^6.0.1",
     "rollup": "^4.20.0",
     "typescript": "~5.5.4",
-    "vitest": "1.2.2",
+    "vitest": "^2.0.5",
```
ripecosta, line 120: vitest v2 ([changelog](https://github.com/vitest-dev/vitest/releases/tag/v2.0.0)) enables `coverage.ignoreEmptyLines` by default so coverage is expected to drop slightly in some files

marcalexiei (commented):
IMHO I think it would be better if `react` and `@types/react` have a more recent version, I would go with `v18`.
This means that we can use more recent version of other test dependencies, as stated in @ripecosta's comments.

What do you think @adrai?

test/typescript/misc/Trans.test.tsx
marcalexiei: This assertion is necessary since the test checks that an error is thrown when there are object inside child computations and `allowObjectInHTMLChildren` is disabled. 

Check the line below:

```tsx
<span>This {{ var: '' }} is an error since `allowObjectInHTMLChildren` is disabled</span>
```

---

Probably that wasn't an issue with old react types. 
This is another reason to update `react` version rather than downgrading `@types/react`

adrai:
> IMHO I think it would be better if `react` and `@types/react` have a more recent version, I would go with `v18`.
> 
> This means that we can use more recent version of other test dependencies, as stated in @ripecosta's comments.
> 
> 
> 
> What do you think @adrai?
> 
> 

Yes, we can update to v18

ripecosta:
Sounds good @adrai @marcalexiei. In that case for this PR I'll keep @testing-library/react-hooks at v7 so we can upgrade react in it's own PR. I will require releasing a new major version too so it's another reason to keep it separate.

test/typescript/misc/Trans.test.tsx
ripecosta: Reverted the `@types/react` downgrade since we decided to upgrade react instead.

PR #1789, opened 2024-08-14
feat: upgrade to react v18

ripecosta:
### Description: 

Upgrades to react v18.

#### Checklist

- [x] tests are included


test/useTranslation.loadingLng.spec.js
```diff
@@ -23,14 +23,16 @@ describe('useTranslation loading ns with lng via props', () => {
...
-    );
-    expect(result.all).toHaveLength(0);
+    const all = [];
+    const { result } = renderHook(() => {
+      const value = useTranslation('common', { i18n: newI18n, useSuspense: true, lng: 'de' });
+      all.push(value);
+      return value;
+    });
+
+    expect(all).toHaveLength(0);
```
ripecosta, line 33: This expectation isn't doing much here. I opted for leaving it as close to what it was as possible but I think the test could be simplified by removing this assertion as well as the line and the `const all`.

adrai:
Theoretically, if we keep the peerDependency of react as is, there is no need for a breaking release, since the v18 react is only needed because of the tests.
I know there are a lot of users still using react 16/17...
I would prefer not to drop this older react compatibility for now.
We can say older versions are not supported, but technically still compatible.
What do you think?

ripecosta:
@adrai If in the future we use a feature that isn't compatible with v16/17 we could that break compatibility without realising. However since the project is pretty stable (and so is react) I guess that's not very likely to happen.  I'm happy to not bump the react version if that's your preference.

Any thoughts on the test with the comment? Leave as is/simplify?

adrai:
ok, 👍 
Regarding the test: it is important in that test case that no render is triggered.

ripecosta:
I see. In that case I'll leave the test as is.

I have reverted the change to the react peer dep so the PR should be ready to merge.

adrai:
thank you 👍

PR #1806, opened 2024-10-21
fix:  `<Trans />` warns 'Each child in a list should have a unique "key" prop.'

aiktb:
close https://github.com/i18next/react-i18next/issues/1805

#### Checklist

- [x] only relevant code is changed (make a diff before you submit the PR)
- [x] run tests `npm run test`
- [ ] tests are included
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/i18next/.github/blob/master/CONTRIBUTING.md)

#### Checklist (for documentation change)

- [ ] only relevant documentation part is changed (make a diff before you submit the PR)
- [ ] motivation/reason is provided
- [ ] commit message and code follows the [Developer's Certification of Origin](https://github.com/i18next/.github/blob/master/CONTRIBUTING.md)

Issue #1805, opened 2024-10-20
[react@19] Using `<Trans />` with the `components` prop gives an error 'Each child in a list should have a unique "key" prop.'


aiktb:
## 🐛 Bug Report

When using **_react@19_**, the `Trans` component throws an error. It has been confirmed that react@18 does not have this issue, and this error does not affect the normal operation of i18next.

This issue has a low priority because react@19 has not been officially released yet.
![image](https://github.com/user-attachments/assets/3b099ad9-690a-4830-afa7-0b44ca19eba7)

## To Reproduce

https://stackblitz.com/edit/vitejs-vite-ufa4wi?file=src%2FApp.tsx,package.json

```bash
npm install react@rc react-dom@rc
```
The sample code is from the react-i18next document:
```jsx
    <Trans
      i18nKey="myKey" // optional -> fallbacks to defaults if not provided
      defaults="hello <italic>beautiful</italic> <bold>{{what}}</bold>" // optional defaultValue
      values={{ what: 'world' }}
      components={{ italic: <i />, bold: <strong /> }}
    />
```
This error has been known to be eliminated using `<i key="1"/>`.

## Expected behavior

No error.

## Your Environment

```txt
    "i18next": "^23.16.1",
    "react-i18next": "^15.0.3",
    "react": "^19.0.0-rc-65a56d0e-20241020",
    "react-dom": "^19.0.0-rc-65a56d0e-20241020",
```

adrai:
it's just a "warning"...
if you like you can try to provide a PR that adds a key for the components, somewhere here: https://github.com/i18next/react-i18next/blob/master/src/TransWithoutContext.js#L365

aiktb:
> it's just a "warning"... if you like you can try to provide a PR that adds a key for the components, somewhere here: https://github.com/i18next/react-i18next/blob/master/src/TransWithoutContext.js#L365

```diff
    Object.keys(components).forEach((c) => {
+     components[c] = cloneElement(components[c], { key: c });
      const comp = components[c];
```

I tested this locally and it removes the warning, is this correct?

adrai:
Maybe better like this?

```js
Object.keys(components).forEach((c, i) => {
      if (!components[c].key) components[c] = cloneElement(components[c], { key: i });
      const comp = components[c];
```

aiktb:
> Maybe better like this?
> 
> ```js
> Object.keys(components).forEach((c, i) => {
[...]

PR #1813, opened 2024-11-22
fix: Attempted to assign to readonly property

ahmadfsalameh:
#### Checklist

- [ ] tests are included

### DIFF
--- a/src/TransWithoutContext.js
+++ b/src/TransWithoutContext.js
@@ -362,8 +362,8 @@ export function Trans({
 
   if (components) {
     Object.keys(components).forEach((c) => {
-      if (!components[c].key) components[c] = cloneElement(components[c], { key: c });
-      const comp = components[c];
+      const componentKey = components[c].key || c;
+      const comp  = cloneElement(components[c], { key: componentKey });
       if (
         typeof comp.type === 'function' ||
         !comp.props ||


adrai:
Can you provide the original issue for this PR?
Or at least describe with more detail information what this PR exactly should fix?

adrai:
> Can you provide the original issue for this PR? Or at least describe with more detail information what this PR exactly should fix?

@ahmadfsalameh can you provide more information please?

ahmadfsalameh:
@adrai A few of our users got this error. When I investigated the stack trace, it led me to https://github.com/i18next/react-i18next/blob/master/src/TransWithoutContext.js#L365, where a recent pull request (https://github.com/i18next/react-i18next/pull/1806) introduced code that mutates the components passed in the props to add missing keys. This is an incorrect approach according to the [React docs](https://react.dev/reference/react/createElement#caveats:~:text=You%20must%20treat%20React%20elements%20and%20their%20props%20as%20immutable%20and%20never%20change%20their%20contents%20after%20creation.%20In%20development%2C%20React%20will%20freeze%20the%20returned%20element%20and%20its%20props%20property%20shallowly%20to%20enforce%20this.).

> You must treat React elements and their props as [immutable](https://en.wikipedia.org/wiki/Immutable_object) and never change their contents after creation. In development, React will [freeze](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze) the returned element and its props property shallowly to enforce this.

Screenshot 2024-11-25 at 3 26 20 PM

aiktb:
The changes in this PR are wrong. `comp` is just a temporary variable abbreviation. It makes no sense to assign it to `cloneElement` and add a key, so it will definitely lead to #1805 reopen.

I cannot reproduce the problem mentioned in this PR using `react@19` and `react-i18next@15.1.1`

aiktb:
@adrai I also noticed that [react does freeze props in development mode](https://github.com/facebook/react/blob/7283a213dbbc31029e65005276f12202558558fc/packages/react/src/jsx/ReactJSXElement.js#L279), but it's a shallow freeze, you can even run the following code, so what does this PR mention have to do with the changes in #1805? idk

```js
const obj = {
    props: {
        comp: {}
    }
};
Object.freeze(obj.props);
obj.props.comp.key = "value";

console.log(obj.props.comp.key === "value");
```

adrai:
@aiktb can you provide a new PR that reverts this...? if @ahmadfsalameh can create a new PR if he founds a better way to address his issue.

aiktb:
> @aiktb can you provide a new PR that reverts this...? if @ahmadfsalameh can create a new PR if he founds a better way to address his issue.

My network environment is very bad now, and I can do it when I wake up in 8 hours

Issue #1815, opened 2024-11-28
Self-closing REACT components in translation strings should not attempt to replace the component's children


simkentys:
## 🐛 Bug Report

When using React components with children to replace self closing tags in translations, the children of react component is replaced.
Probably continuation of this issue: https://github.com/i18next/react-i18next/issues/1695

## To Reproduce

https://stackblitz.com/edit/vitejs-vite-f6qrql?file=package.json,src%2Fmain.tsx%3AL41&terminal=dev

```js
i18n.use(initReactI18next).init({
  interpolation: {
    escapeValue: false,
  },
});

function Link({ href, children }) {
  return <a href={href}>{children}</a>;
}

createRoot(document.getElementById('root')!).render(
  <>
      <Link href="#">linkTo</Link>
      <Trans
        i18nKey="<Component />"
        components={{ Component: <Link href="#">linkTo</Link> }}
      />
      <Trans
        i18nKey="<Component />"
        components={{ Component: <a href="#">linkTo</a> }}
      />
  </>
);
```

expected result:
```html
<a href="#">linkTo</a>
<a href="#">linkTo</a>
<a href="#">linkTo</a>
```
actual result:
```html
<a href="#">linkTo</a>
<a href="#"></a>
<a href="#">linkTo</a>
```

## Expected behavior

React component children would be preserved

## Your Environment

"i18next": "^24.0.2",
"react-i18next": "^15.1.2"

adrai:
Feel free to provide a PR to address this.

adrai:
addressed in v15.1.3

PR #1816, opened 2024-11-29
fix issue 1815

simkentys:
fix issue #1815

#### Checklist

- [X] tests are included
- [X] commit message and code follows the [Developer's Certification of Origin](https://github.com/i18next/.github/blob/master/CONTRIBUTING.md)

### DIFF
--- a/src/TransWithoutContext.js
+++ b/src/TransWithoutContext.js
@@ -365,7 +365,6 @@ export function Trans({
       const componentKey = components[c].key || c;
       const comp = cloneElement(components[c], { key: componentKey });
       if (
-        typeof comp.type === 'function' ||
         !comp.props ||
         !comp.props.children ||
         (translation.indexOf(`${c}/>`) < 0 && translation.indexOf(`${c} />`) < 0)
--- a/test/trans.render.spec.jsx
+++ b/test/trans.render.spec.jsx
@@ -648,6 +648,70 @@ describe('trans should work with selfclosing elements in components', () => {
   });
 });
 
+describe('trans should work with self closing elements with react components', () => {
+  function Component({ children }) {
+    return <div>{children}</div>;
+  }
+  it('should render component children', () => {
+    const { container } = render(
+      <Trans
+        i18nKey="transTestWithSelfClosing"
+        components={{
+          component: <Component>These children will be preserved</Component>,
+        }}
+      />,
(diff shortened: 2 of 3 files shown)


src/TransWithoutContext.js
```diff
@@ -3,6 +3,11 @@ import HTML from 'html-parse-stringify';
 import { isObject, isString, warn, warnOnce } from './utils.js';
 import { getDefaults } from './defaults.js';
 import { getI18n } from './i18nInstance.js';
+import { Fragment, isValidElement, cloneElement, createElement, Children } from 'react';
```
adrai, line 6: Why are you importing these again? it's already imported above

test/trans.render.spec.jsx
```diff
@@ -648,6 +648,30 @@ describe('trans should work with selfclosing elements in components', () => {
   });
 });
 
+describe('trans should work with self closing elements with react components', () => {
```
adrai, line 651: can you add another test with your examples? <Link> and <a>

src/TransWithoutContext.js
simkentys: my bad, not sure what happened here. reset the commits and tried again, it seems that some process adds these during the commit, could it be some git hooks?

adrai: there is no such magic...
but also remove this and try again: "trim_trailing_whitespace = true"

simkentys: tried with -n option, still the same, guessing some issue with my ide. About trim whitespaces, if it's enabled, whenever you edit tests the trailing spaces gets removed in html elements, and a lot of tests fail because of that, should i still enable it?

adrai: this does not happen on my machine

simkentys: probably intellij issue then, ill restore it

simkentys: fixed imports, and restore trailing spaces

test/trans.render.spec.jsx
simkentys: what other test should i add?

simkentys: added more tests

adrai:
thank you, included in v15.1.3

Issue #1733, opened 2024-03-08
`Trans` component should NOT modify the input props value.


Jayatubi:
I've got an issue around this line in `TransWithoutContext.js`:

https://github.com/i18next/react-i18next/blob/b274be772bde64377ce1d75f28c7cd3f46735851/src/TransWithoutContext.js#L380

```javascript
  if (components) {
    Object.keys(components).forEach((c) => {
      const comp = components[c];
      if (
        typeof comp.type === 'function' ||
        !comp.props ||
        !comp.props.children ||
        (translation.indexOf(`${c}/>`) < 0 && translation.indexOf(`${c} />`) < 0)
      )
        return;

      // eslint-disable-next-line react/no-unstable-nested-components, no-inner-declarations
      function Componentized() {
        // <>{comp}</>
        return createElement(Fragment, null, comp);
      }
      // <Componentized />
      components[c] = createElement(Componentized);
      // ^ `components` may be frozen and then crash happens
    });
  }
```

Looks like `Trans` was going to modify the input prop `components` inplace. However, this may lead a crash while the `components`  is `readonly`, aka `frozen`.

I encounter this issue by accidentally storing a React component with Trans by using `useImmer` hook. The `useImmer` hook will freeze all the states in production environment by default.

For exmaple: 
```javascript
const MyComponent = () => {
    // The components props will be frozen in production environment
    const [Inner, setInner] = useImmer(<>
        <Trans components={components}></Trans>
    </>)

    return <Inner/>;
}
```

So instead of using `forEach` to convert the input `components` may be it would be better to use `map` to create a new variable for later use.

adrai:
Feel free to provide a PR to address this.

PR #1820, opened 2024-12-07
Fix: warning each child should have a unique key

ahmadfsalameh:
#### Checklist

- [ ] tests are included

@adrai @aiktb @XAOPT When I made the changes in https://github.com/i18next/react-i18next/pull/1813 I wasn't aware of the conditional return, It seems in the example provided here: https://stackblitz.com/edit/vitejs-vite-59g4bx?file=package.json the first [if statement](https://github.com/i18next/react-i18next/blob/master/src/TransWithoutContext.js#L367-L372) in `TransWithoutContext` is returning void causing the unmodified components (without key) to get rendered.

That function handles both arrays and objects using `Object.keys` which is smart but also very confusing. We could revert my changes and the warning will disappear but it will cause errors to show up because those components are readonly, also as specified here in your type: https://github.com/i18next/react-i18next/blob/master/TransWithoutContext.d.ts#L16.

In this pull request I'm attempting to solve both issues, mutating the component prop and the keys warning. I split the function into smaller functions which will return a new components value at the end without overriding it.

It might also fix https://github.com/i18next/react-i18next/issues/1733

### DIFF
--- a/src/TransWithoutContext.js
+++ b/src/TransWithoutContext.js
@@ -298,6 +298,59 @@ const renderNodes = (children, targetString, i18n, i18nOptions, combinedTOpts, s
   return getChildren(result[0]);
 };
 
+const fixComponentProps = (component, index, translation) => {
+  const componentKey = component.key || index;
+  const comp = cloneElement(component, { key: componentKey });
+  if (
+    !comp.props ||
+    !comp.props.children ||
+    (translation.indexOf(`${index}/>`) < 0 && translation.indexOf(`${index} />`) < 0)
+  ) {
+    return comp;
+  }
+
+  function Componentized() {
+    // <>{comp}</>
+    return createElement(Fragment, null, comp);
+  }
+  // <Componentized />
+  return createElement(Componentized);
+};
+
+const generateArrayComponents = (components, translation) =>
+  components.map((c, index) => fixComponentProps(c, index, translation));
+
+const generateObjectComponents = (components, translation) => {
+  const componentMap = {};
+
+  Object.keys(components).forEach((c) => {
+    Object.assign(componentMap, {
+      [c]: fixComponentProps(components[c], c, translation),
+    });
+  });
+
+  return componentMap;
+};
+
+const generateComponents = (components, translation) => {
+  if (!components) return null;
+
+  // components could be either an array or an object
+
+  if (Array.isArray(components)) {
+    return generateArrayComponents(components, translation);
+  }
+
+  if (isObject(components)) {
+    return generateObjectComponents(components, translation);
+  }
+
+  // if components is not an array or an object, warn the user
+  // and return null
+  console.warn('<Trans /> component prop expects an object or an array');
+  return null;
+};
+
 export function Trans({
   children,
   count,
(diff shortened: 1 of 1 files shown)


src/TransWithoutContext.js
```diff
@@ -298,6 +298,59 @@ const renderNodes = (children, targetString, i18n, i18nOptions, combinedTOpts, s
...
+    return generateArrayComponents(components, translation);
+  }
+
+  if (isObject(components)) {
+    return generateObjectComponents(components, translation);
+  }
+
+  // if components is not an array or an object, warn the user
+  // and return null
+  console.warn('<Trans /> component prop expects an object or an array');
```
adrai, line 350: Please use warnOnce here: https://github.com/i18next/react-i18next/blob/master/src/utils.js#L9

aiktb:
This looks ok, I tested it locally with the patched code and it successfully fixed the warning.

aiktb:
Can you add a test for this? @adrai This does not need to be done in this PR
I tried upgrading the react-related development dependencies to 19, which did not break any tests, but some peer dependencies were broken. Upgrading the following dependencies eliminated the peer dependency errors and all tests passed:

```diff
- "@testing-library/react": "^16.0.0",
+ "@testing-library/react": "^16.1.0",

- "@types/react": "^18.3",
+ "@types/react": "^19.0.0",

- "react": "^18.3.1",
+ "react": "^19.0.0",

- "react-dom": "^18.3.1",
+ "react-dom": "^19.0.0",

- "react-test-renderer": "^18.3.1",
+ "react-test-renderer": "^19.0.0",
```

adrai:
> Can you add a test for this? @adrai This does not need to be done in this PR
> 
> I tried upgrading the react-related development dependencies to 19, which did not break any tests, but some peer dependencies were broken. Upgrading the following dependencies eliminated the peer dependency errors and all tests passed:
> 
> 
> 
> ```bash
> 
> npm install @testing-library/react@^16.1.0 @types/react@^19.0.0 react@^19.0.0 react-dom@^19.0.0 react-test-renderer@^19.0.0
> 
> ```
> 
> 
> 
> ```diff
> 
> - "@testing-library/react": "^16.0.0",
> 
> + "@testing-library/react": "^16.1.0",
> 
> 
> 
> - "@types/react": "^18.3",
> 
> + "@types/react": "^19.0.0",
> 
> 
> 
> - "react": "^18.3.1",
> 
> + "react": "^19.0.0",
> 
> 
> 
> - "react-dom": "^18.3.1",
> 
> + "react-dom": "^19.0.0",
> 
> 
> 
> - "react-test-renderer": "^18.3.1",
> 
> + "react-test-renderer": "^19.0.0",
> 
> ```

Can you create a PR for the deps update?

aiktb:
> > Can you add a test for this? @adrai This does not need to be done in this PR
> > I tried upgrading the react-related development dependencies to 19, which did not break any tests, but some peer dependencies were broken. Upgrading the following dependencies eliminated the peer dependency errors and all tests passed:
> > ```shell
> > npm install @testing-library/react@^16.1.0 @types/react@^19.0.0 react@^19.0.0 react-dom@^19.0.0 react-test-renderer@^19.0.0
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```diff
> > - "@testing-library/react": "^16.0.0",
> > 
> > + "@testing-library/react": "^16.1.0",
> > 
> > 
> > 
> > - "@types/react": "^18.3",
> > 
> > + "@types/react": "^19.0.0",
> > 
> > 
> > 
> > - "react": "^18.3.1",
> > 
> > + "react": "^19.0.0",
> > 
> > 
> > 
> > - "react-dom": "^18.3.1",
> > 
> > + "react-dom": "^19.0.0",
> > 
> > 
> > 
> > - "react-test-renderer": "^18.3.1",
> > 
> > + "react-test-renderer": "^19.0.0",
> > ```
> 
> Can you create a PR for the deps update?

https://github.com/i18next/react-i18next/pull/1821

src/TransWithoutContext.js
ahmadfsalameh: Sure! 815ff0ffa578c8bbf528bc070f84e6272c8909af

adrai:
> Can you add a test for this?

@ahmadfsalameh 👆

adrai:
merged and released with v15.1.4... please @ahmadfsalameh if possible create a new PR with a dedicated test for this

PR #1826, opened 2024-12-29
feat:  add meta with codes on warnings to allow conditional logging

delta-9:
#### Checklist

- [ ] tests are included

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -1,20 +1,22 @@
-export const warn = (i18n, ...args) => {
+/** @type {(i18n:any,code:import('../TransWithoutContext').ErrorCode,msg?:string, rest?:{[key:string]: any})=>void} */
+export const warn = (i18n, code, msg, rest) => {
+  const args = [msg, { code, ...(rest || {}) }];
   if (i18n?.services?.logger?.forward) {
-    i18n.services.logger.forward(args, 'warn', 'react-i18next::', true);
-  } else if (i18n?.services?.logger?.warn) {
-    if (isString(args[0])) args[0] = `react-i18next:: ${args[0]}`;
+    return i18n.services.logger.forward(args, 'warn', 'react-i18next::', true);
+  }
+  if (isString(args[0])) args[0] = `react-i18next:: ${args[0]}`;
+  if (i18n?.services?.logger?.warn) {
     i18n.services.logger.warn(...args);
   } else if (console?.warn) {
-    if (isString(args[0])) args[0] = `react-i18next:: ${args[0]}`;
     console.warn(...args);
   }
 };
-
 const alreadyWarned = {};
-export const warnOnce = (i18n, ...args) => {
-  if (isString(args[0]) && alreadyWarned[args[0]]) return;
-  if (isString(args[0])) alreadyWarned[args[0]] = new Date();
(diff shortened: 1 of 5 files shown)


adrai:
not necessary, since you can now inject your own logger: https://github.com/i18next/react-i18next-gitbook/pull/150#issuecomment-2564710238

delta-9:
@adrai I saw your changes just now, cool!, this PR now just adds meta so I can filter what to log or not

adrai:
I suggest, you move this logic out to your custom logger... else this will bloat the code size even more... //cc @VIKTORVAV99

delta-9:
> I suggest, you move this logic out to your custom logger... else this will bloat the code size even more... //cc @VIKTORVAV99

fair enough, I reworked the changes to includes only what is needed to do that from the logger

adrai:
Do you really need extra error codes?

delta-9:
> Do you really need extra error codes? The whole i18next core code base has no error codes...

the error codes is to be able to identify the warnings in the code.
it also adds the i18nkey to the warning logged so its easier to find the culprit trans.

the i18nkey also allow to add a logic to warnOnce per key in the custom logger, now it warn on each renders.

for the code I could do some string regexp on the messages to figure out the cause of the error,
and for the i18nkey, Its also possible to debug following the stack trace and put some trace until I find the right call.

VIKTORVAV99:
Maybe there are benefits from this but I would like you to consider two things:

First this will increase the overhead and bundle size for all users of i18next, I've been trying to push this bundle down as it's one of the bigger parts of our app we can't change at the moment.

Second, I would almost call this a breaking change, partly as it changes the log outputs which means all tests or error identifying software such as sentry will either fail or log these as new errors. Secondly it feels like this is just duplicating the error message. If we do this I would switch to something what react does and only log the error codes with a link to look up what they mean. But that requires other changes as well.

That is just my thought on the matter though but I would really like to avoid duplicating any information or functionality for the sake of making it a little more convenient for one project or developer.

rawpixel-vincent:
I replaced the error map with a tuple, and reduced duplicates, it's adding 201 bytes to `react-i18next.min.js` build compared to current master

src/utils.js
```diff
@@ -1,20 +1,33 @@
-export const warn = (i18n, ...args) => {
+/** @template {string} T @param {...T} s @return {T[]} */
+const tup = (...s) => s;
```
adrai, line 2: The whole ERR_CODES and tup code can be removed now, right?

delta-9:
If you ask me, I think it's needed for sanity.
But the code can run without it and the type accept any string as code.

delta-9:
good news @adrai I moved the error code to a type declaration file so its not in the code.

the minify js is now 20 bytes less than master 🎉

delta-9:
added types for along an example of usage in the logger plugin

adrai:
TS tests seems to fail: https://github.com/i18next/react-i18next/actions/runs/12542746999/job/34976361013?pr=1826

delta-9:
it should be fixed, but I can't make the test fail locally, npm run test pass

adrai:
it's included in v15.4.0

PR #1842, opened 2025-04-22
feat: use const type parameters for useTranslation()

emersion:
TypeScript 5.0 has introduced [const type parameters][1]. These indicate that a more specific type is preferred over a more general one. As a result, the resulting UseTranslationResponse will be more precise and retain the passed-in namespace and prefix. This can be used by extractors and other tooling to find out exactly which key is accessed by a translation function call.

[1]: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-0.html#const-type-parameters

#### Checklist

- [x] tests are included (N/A: there is no way to extract generic type parameters from a test)

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -97,8 +97,8 @@ export type FallbackNs<Ns> = Ns extends undefined
     : _DefaultNamespace;
 
 export function useTranslation<
-  Ns extends FlatNamespace | $Tuple<FlatNamespace> | undefined = undefined,
-  KPrefix extends KeyPrefix<FallbackNs<Ns>> = undefined,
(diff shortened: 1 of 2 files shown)


adrai:
@marcalexiei ok for you?

marcalexiei (commented):
Since you are introducing TypeScript 5 features consider to make it an optional dependency.
It should be the same of `i18next`:

https://github.com/i18next/i18next/blob/c40d0a2744582deedacf7d10e534182449992e4e/package.json#L54-L61

emersion:
Makes sense to me, done!

emersion:
Thanks for the speedy review!

adrai:
thanks both of you...
included in v15.5.0

PR #1854, opened 2025-07-03
Fix: passing components as object should still allow for indexed matching of children

FritsvanCampen:
I encountered what I think is a bug.

Consider the following case:

```
<Trans i18nKey="tkey" default="Click <1>here</1> to continue">
  pre <a href="#somewhere">link</a> post
</Trans>
```

This outputs `Click [here](#somewhere) to continue` as expected. However I expect that it keeps working even if I pass in a components map that isn't actually used:

```
<Trans i18nKey="tkey" default="Click <1>here</1> to continue" components={{ unused: <span /> }}>
  pre <a href="#somewhere">link</a> post
</Trans>
```

Before my change this would no longer match `<1>` to the `<a>` passed in children, because children was being ignored by the components map. Part of the code wrongfully assumes that `components` contains something like `{ 1: <a href="#somewhere> }`, but it doesn't in this case; I'm using `components` as a map, not an index - which is also definitely supported!

I struggled a bit with the code because `children` is very polymorphic, the code has poorly named variables, and logic is all over the place; luckily the tests are good so I'm confident this doesn't break any other use case 😅

#### Checklist

- [ x ] only relevant code is changed (make a diff before you submit the PR)
- [ x ] run tests `npm run test`
- [ x ] tests are included
- [ x ] commit message and code follows the [Developer's Certification of Origin](https://github.com/i18next/.github/blob/master/CONTRIBUTING.md)

### DIFF
--- a/src/TransWithoutContext.js
+++ b/src/TransWithoutContext.js
@@ -113,7 +113,15 @@ export const nodesToString = (children, i18nOptions, i18n, i18nKey) => {
   return stringNode;
 };
 
-const renderNodes = (children, targetString, i18n, i18nOptions, combinedTOpts, shouldUnescape) => {
+const renderNodes = (
+  children,
+  knownComponentsMap,
+  targetString,
+  i18n,
+  i18nOptions,
+  combinedTOpts,
+  shouldUnescape,
+) => {
   if (targetString === '') return [];
 
   // check if contains tags we need to replace from html string to react nodes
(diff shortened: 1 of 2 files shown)


FritsvanCampen:
I see codeclimate has found an issue. I'm not sure if I need to fix that, the condition it complains about isn't actually that complicated and not more complicated than existing conditions in the code. Perhaps codeclimate is a little over-tuned? 🤷

adrai:
ignore code climate

adrai:
Your change is very interesting...
If the tests are green this is a good starting point, but it may be there is some non-covered edge-case in the wild...
I would suggest, we merge it and release a new minor version.
If there comes up some strange issue related to this, are you ok with it if we assign thise issues to you?

FritsvanCampen:
That's fine 👍

adrai:
Thanks a lot...
it's included in v15.6.0

FritsvanCampen:
Thank you for the quick response 🙏

PR #1856, opened 2025-07-22
add regression test for optional chaining issue

bjoerge:
According to the [types](https://github.com/i18next/react-i18next/blob/master/index.d.ts#L74), calling `useTranslation(ns, {bindI18n: false})` is valid and supported. However, this currently will crash at runtime with the error: 
```
TypeError: options.bindI18n?.indexOf is not a function
```

This happens because the optional chaining here resolves `indexOf` to `undefined`, which is then invoked:

https://github.com/i18next/react-i18next/blob/d515c68a4bf2280b5248dd3ab050b62b59fe98c0/src/utils.js#L75

See regression test demonstrating the issue in 4dfded7. I also included a test for explicitly passing `undefined` – (which means it won't be overridden by default options), not sure how valuable this is, but figured there's little harm in including it.

See b9381e6 for the fix

This is currently blocking us from upgrading from 14.1.3 to 15.6.0 (see https://github.com/sanity-io/sanity/pull/10050)

#### Checklist

- [x] tests are included

### DIFF
--- a/src/useTranslation.js
+++ b/src/useTranslation.js
@@ -132,7 +132,7 @@ export const useTranslation = (ns, props = {}) => {
     // unbinding on unmount
     return () => {
       isMounted.current = false;
-      if (i18n) bindI18n?.split(' ').forEach((e) => i18n.off(e, boundReset));
+      if (i18n && bindI18n) bindI18n?.split(' ').forEach((e) => i18n.off(e, boundReset));
       if (bindI18nStore && i18n)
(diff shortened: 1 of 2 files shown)


bjoerge:
whoops, looks like the PR title didn't reflect that it got fixed, sorry about that. But thanks for merging @adrai!

adrai:
I have to thank you....

included in v15.6.1

bjoerge:
Amazing, thank you so much!

Issue #1860, opened 2025-08-28
Generic type 'TransProps' requires between 1 and 6 type arguments


souloffice:
## Documentation issue
Generic type 'TransProps' requires between 1 and 6 type arguments

Not able to understand how to extends TransProps of react-i18next to create CustomTrans Component.

[Repo to reproduce](https://github.com/souloffice/react-i18next-issue)

adrai:
originally: https://github.com/i18next/next-i18next/issues/2325#issuecomment-3227761924

I'm not a TypeScript expert... but maybe like this?

```ts
import { type TransProps, Trans } from "react-i18next";
import type { ParseKeys, Namespace, TypeOptions, TOptions } from "i18next";

type _DefaultNamespace = TypeOptions['defaultNS'];

type CustomTransProps<
  Key extends ParseKeys<Ns, TOpt, KPrefix>,
  Ns extends Namespace = _DefaultNamespace,
  KPrefix = undefined,
  TContext extends string | undefined = undefined,
  TOpt extends TOptions & { context?: TContext } = { context: TContext },
  E = React.HTMLProps<HTMLDivElement>,
> = TransProps<Key, Ns, KPrefix, TContext, TOpt, E> & {
  className?: string;
}

export function CustomTrans<
  Key extends ParseKeys<Ns, TOpt, KPrefix>,
  Ns extends Namespace = _DefaultNamespace,
  KPrefix = undefined,
  TContext extends string | undefined = undefined,
  TOpt extends TOptions & { context?: TContext } = { context: TContext },
  E = React.HTMLProps<HTMLDivElement>,
>(
  props: CustomTransProps<Key, Ns, KPrefix, TContext, TOpt, E>
): React.ReactElement {
  const { className } = props;

  return <Trans className={className} {...props} />;
}
```

@ahrjarrett is this correct?
And how would the new selector API be used here with the CustomTrans?
// cc @marcalexiei

ahrjarrett:
@souloffice I created a [Playground (outdated)](https://tsplay.dev/wQax1w) to demonstrate how you could wrap the `Trans` component for both APIs.

> [!WARNING]
> The Playground link above is outdated, please refer to [this one](https://tsplay.dev/wEKRym) instead

While working on that, I realized that I forgot to export `TransSelectorProps` from `react-i18next`, so the selector API example won't work until #1862 is merged.

Let me know if you have any questions, or if that doesn't work for your use case!

adrai:
[...]

PR #1862, opened 2025-08-28
chore: exports `TransSelectorProps`

### LINKED ISSUE #1861: `TransSelectorProps` is not exported
## 🐛 Bug Report

As I was looking into #1860, I realized that `TransSelectorProps` is not exported from `index.d.ts`.

We'll need to export that if users are wanting to wrap the `Trans` component.

## To Reproduce

[Playground](https://tsplay.dev/wQax1w)

```js
import type { TransSelectorProps } from 'react-i18next' // 🚫
```

## Expected behavior

To be able to import the `TransSelectorProps` type

```js
import type { TransSelectorProps } from 'react-i18next' // ✅
```

ahrjarrett:
Closes #1861 
Related to #1860

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -10,11 +10,17 @@ import type {
   KeyPrefix,
 } from 'i18next';
 import * as React from 'react';
-import { Trans, TransProps, ErrorCode, ErrorArgs } from './TransWithoutContext.js';
+import {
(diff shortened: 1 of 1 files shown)


adrai:
thx a lot, included in v15.7.3

PR #1868, opened 2025-10-13
fix: check to make sure translation key is a string before attempting to translate it

### LINKED ISSUE #1867: Using `<Trans>` component with named tags throws error when Selector API is enabled
## 🐛 Bug Report

When using `Trans` with the Selector API enabled, throws errors when components passed via the `components` prop 

Image

## To Reproduce

```
<Trans
    t={t}
    i18nKey={($) => $['testString']}
    values={{ buttonText: 'button' }}
    components={{
        el1: <a href="https://www.google.com/">Link</a>,
        el2: (
            <Button variant="link" onClick={() => console.log('clicked')}>
                Button
            </Button>
        ),
    }}
/>
```
with translation key:

```js
{
    "testString": "This has an interpolated <el1>link</el1> and <el2>%{buttonText}</el2>"
}
```

## Expected behavior

Should not throw an error.

## Your Environment

- *runtime version*: v22.18.0
- *i18next version*: v25.5.3
- *react-i18next version*: v15.7.1
- *os*: Mac

ahrjarrett:
Fixes #1867

This PR changes how the `defaultValue` is resolved in the `Trans` component.

The reason for the crash is because `defaultValue` was falling back to whatever was passed as the `i18nKey`, which in this case was a selector function.

Then, because no `lng` was specified (and `defaultValue` is added to [`combinedTOpts`](https://github.com/i18next/react-i18next/blob/master/src/TransWithoutContext.js#L451)), the [`translation`](https://github.com/i18next/react-i18next/blob/master/src/TransWithoutContext.js#L454) ends up falling back to the `defaultValue` (which is still a selector function), which eventually blows up in [`fixComponentProps`](https://github.com/i18next/react-i18next/blob/master/src/TransWithoutContext.js#L327).

To fix this, we now check if `i18nKey` is a function, and if so, we pass it to `keysFromSelector` to turn the selector function into a string key.

#### Checklist

- [x] tests are included

### DIFF
--- a/src/TransWithoutContext.js
+++ b/src/TransWithoutContext.js
@@ -415,9 +415,6 @@ export function Trans({
     return children;
   }
 
-  // eslint-disable-next-line no-param-reassign
-  if (typeof i18nKey === 'function') i18nKey = keyFromSelector(i18nKey);
-
   const t = tFromProps || i18n.t.bind(i18n) || ((k) => k);
 
   const reactI18nextOptions = { ...getDefaults(), ...i18n.options?.react };
@@ -428,7 +425,10 @@ export function Trans({
 
   const nodeAsString = nodesToString(children, reactI18nextOptions, i18n, i18nKey);
   const defaultValue =
-    defaults || nodeAsString || reactI18nextOptions.transEmptyNodeValue || i18nKey;
+    defaults ||
+    nodeAsString ||
+    reactI18nextOptions.transEmptyNodeValue ||
+    (typeof i18nKey === 'function' ? keyFromSelector(i18nKey) : i18nKey);
   const { hashTransKey } = reactI18nextOptions;
   const key =
     i18nKey ||


adrai:
still afk...
Without a lng there should be logged a warning somewhere, but no crash...

ahrjarrett:
> Without a lng there should be logged a warning somewhere, but no crash...

Agreed. This PR fixes the crashing behavior.

The reason it was crashing is because `defaultValue` was falling back to whatever was passed as the `i18nKey`, which in this case was a selector function.

Then, because no `lng` was specified (and `defaultValue` is added to [`combinedTOpts`](https://github.com/i18next/react-i18next/blob/master/src/TransWithoutContext.js#L451)), the [`translation`](https://github.com/i18next/react-i18next/blob/master/src/TransWithoutContext.js#L454) ends up falling back to the `defaultValue` (the selector function), which eventually blows up in [`fixComponentProps`](https://github.com/i18next/react-i18next/blob/master/src/TransWithoutContext.js#L327).

adrai:
thx

adrai:
included in v16.0.1

Issue #1869, opened 2025-10-13
More robust icu-specific Trans babel macro


cellog:
## 🚀 Feature Proposal

A new `Trans` component that does not rely upon React.children under the hood will allow for more future-compatible work. The current component doesn't actually work with deeply nested trees like this:

```tsx
    return (
      <div>
        <Trans i18nKey="complex-trans">
          <strong>exciting!</strong>
          {plural`${count},
          =0 { hi there ${(<strong>friend</strong>)} }
          other { woweee even supports nested ${number`${numbers}`} } `}{" "}
          and
          {select`${selectInput},
           thing { another nested ${(
             <Box style={{ color: "red" }}>
               with regular text and a date: <code>{date`${now}`}</code>
             </Box>
           )} other {and the fallback <ul><li>one</li><li>two</li></ul>}} `}
        </Trans>
      </div>
    );
```

This renders as:

<img width="442" height="62" alt="Image" src="https://github.com/user-attachments/assets/0dd9d9f5-6f7c-4643-857b-e16abcf75bec" />

I wrote a proof-of-concept new Trans component, and the same JSX renders correctly as:

<img width="356" height="59" alt="Image" src="https://github.com/user-attachments/assets/0a1b45f7-a690-4762-ad07-a975a5e6c5e5" />

The new component is slightly less ergonomic to use. The existing macro compiles to this API:

```tsx
<Trans
  i18nKey="complex-trans"
defaults="<0>exciting!</0>{count, plural, =0 { hi there <1>friend</1> } other { woweee even supports nested {numbers, number} } } and{selectInput, select, thing { another nested <2>with regular text and a date: <0>{now, date}</0></2> } other { fallback }}"
components={[<strong>exciting!</strong>, <strong>friend</strong>, <Box style={{
  color: \\"red\\"
}}>
       with regular text and a date: <code>{date`${now}`}</code>
     </Component>]} values={{
  count,
  numbers,
  selectInput,
  now
}} />
```

Note that it doesn't correctly discover the internal `<code>`.

The new API looks like this (I used `InternalTrans` as the component name, but we can bikeshed that, perhaps `ICUTrans` or something like that):

```tsx
<InternalTrans
  i18nKey="key"
[...]

PR #1870, opened 2025-10-15
Migrate icu macro to typescript

cellog:
This is a preparatory step to adding a new `Trans` component which can be used for the icu macro. It converts the existing icu macro to Typescript, and adds tests to the existing macro. Referencing issue: #1869 

To make things easier to modify later, it splits up the file into many separate files, and adds tests for each specific chunk.

The existing test was moved to typescript, but the snapshot is identical to the old one, showing that it is exactly the same code

#### Checklist

- [x] tests are included

### DIFF
--- a/src/icu.macro/utils/icu-formatter.ts
+++ b/src/icu.macro/utils/icu-formatter.ts
@@ -158,6 +158,7 @@ function processExpressionContainer(
     return formatted;
   }
 
+  // Unknown expression type - return empty string for forward compatibility
   return '';
 }
 
@@ -212,7 +213,8 @@ function processTemplateExpression(
     return { text: formatted, variables: vars };
   }
 
-  return { text: '', variables: [] };
+  // This should never be reached due to validation above
+  throw new Error('Unreachable code in processTemplateExpression');
 }
 
 /**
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -51,10 +51,10 @@
     },
     "./icu.macro": {
       "types": {
-        "require": "./icu.macro..d.ts",
-        "import": "./icu.macro..d.mts"
+        "require": "./icu.macro.d.ts",
+        "import": "./icu.macro.d.mts"
       },
-      "default": "./icu.macro.js"
+      "default": "./dist/commonjs/icu.macro.js"
```
adrai: we may streamline this like the other exports... import, require, etc... commonjs and es

adrai:
Is it really necessary to introduce TypeScript code? With that now we have mixed JS and TS...

adrai:
btw: tests are failing

cellog:
> Is it really necessary to introduce TypeScript code? With that now we have mixed JS and TS...

I am willing to migrate the entire thing to TS, was just trying to limit the scope of this PR. As for why, making a new macro with TS is far safer than making it without. I developed the new component also using TS, which made it far safer.

An alternative approach I am willing to do is to take the component I made and the new macro and strip the typescript and add new javascript in the repo.

package.json
adrai: Since react-i18next was always a JavaScript module and only later shipped with types (but still having javascript code), I would say we stay with JavaScript, until there's a real need to move everything to TypeScript.

PR #1880, opened 2025-10-30
fix trans component break with less than sign

### LINKED ISSUE #1734: Less than sign `<` breaks tag parsing with Trans component
## 🐛 Bug Report

A translation string with less than sign `<` breaks the tag parsing with `<Trans>` component

## To Reproduce

Add this test case to `test/trans.render.object.spec.jsx`: (modifying the test case on line 65 there)
```js
describe('trans using no children but components (object) - using names', () => {
  function TestComponent() {
    return (
      <Trans
        defaults="hello <italic>under ten</italic><10 this text after the sign should be rendered <bold>{{what}}</bold>"
        values={{ what: 'world' }}
      />
    );
  }
    expect(container.firstChild).toMatchInlineSnapshot(`
      <div>
        hello 
        <i>
          under ten
        </i>
        <10 this text after the sign should be redered         
        <strong>
          world
        </strong>
      </div>
    `);
  });
});

```
Run `npm test`, it fails with output
```
 FAIL  |runtime| test/trans.render.object.spec.jsx > trans using no children but components (object) - using names > should render translated string
Error: Snapshot `trans using no children but components (object) - using names > should render translated string 2` mismatched

- Expected
+ Received

  <div>
    hello 
    <i>
      under ten
    </i>
-   <10 this text after the sign should be rendered         
-   <strong>
-     world
-   </strong>
+   world
  </div>

 ❯ test/trans.render.object.spec.jsx:104:34
    102|   it('should render translated string', () => {
    103|     const { container } = render(<TestComponent />);
    104|     expect(container.firstChild).toMatchInlineSnapshot(`
       |                                  ^
    105|       <div>
    106|         hello 

```

## Expected behavior

The test should pass.

The problem is that having a less than sign (`<`) within the translated text breaks the system extracting the tags.

The text between the `<` sign and the next tag opening won't be rendered at all. The next tag won't be rendered correctly.

I assume that in the test case snippet, it reads the following as the tag: `<10 this text after the sign should be rendered <bold>`

## Your Environment

- *runtime version*: node v18.18.0, latest Chrome and Safari
- *i18next version*: `master`, 14.1.0
- *os*: macOS

### LINKED ISSUE #1869: More robust icu-specific Trans babel macro

### LINKED ISSUE #1879: Regression in 16.2.0: `bindI18nStore` does not work correctly
## 🐛 Bug Report

The test below passes on 16.1.4 but fails on 16.2.0: The final assertion expects the text to have changed to German, but it is still in English.

## To Reproduce

Run the following vitest:
```ts
import { act, render, screen } from "@testing-library/react";
import { createInstance } from "i18next";
import { I18nextProvider, useTranslation } from "react-i18next";
import { expect } from "vitest";

it("regression in 16.2.0", async () => {
    const i18next = createInstance();
    await i18next.init({
        fallbackLng: ["en"],
        react: { bindI18nStore: "added" },
    });
    const TranslateAKey = () => {
        const { t } = useTranslation();
        return <h1>{t("key", { ns: "namespace" })}</h1>;
    };
    i18next.addResourceBundle("en", "namespace", { key: "english" }, false, true);
    render(
        <I18nextProvider i18n={i18next}>
            <TranslateAKey />
        </I18nextProvider>,
    );
    expect(screen.getByRole("heading")).toHaveTextContent("english");
    await act(() => i18next.changeLanguage("de"));
    expect(screen.getByRole("heading")).toHaveTextContent("english");
    await act(() => i18next.addResourceBundle("de", "namespace", { key: "deutsch" }, false, true));
    expect(screen.getByRole("heading")).toHaveTextContent("deutsch"); // this assertion fails
});
```

## Expected behavior

The test should pass.

## Your Environment

- *runtime version*: node v22
- *i18next version*: 25.6.0
- Windows
- Relevant packages:
  - "react": "18.3.1"
  - "@testing-library/react": "16.3.0"
  - "i18next": "25.6.0"
  - "react-i18next": "16.2.0"
  - "vitest": "3.2.4"
  - "jsdom": "27.0.1"

codomposer:
## Fixes

Closes #1734 
Closes #1869 
Closes #1879 

#### Checklist

- [x] tests are included

### DIFF
--- a/src/TransWithoutContext.js
+++ b/src/TransWithoutContext.js
@@ -156,8 +156,15 @@ const escapeLiteralLessThan = (str, keepArray = [], knownComponentsMap = {}) => 
       }
 
       // Check for opening tag: <number> or <name> or <name/> or <name />
+      // Also handle tags with attributes: <0 href="..."> or <name class="...">
       if (!isValidTag) {
-        const openingMatch = str.slice(i).match(/^<(\d+|[a-zA-Z][a-zA-Z0-9]*)(\s*\/)?\s*>/);
+        // Match: <tagName [attributes] [/]>
+        // Attributes pattern: name="value" or name='value' or name (boolean)
+        const openingMatch = str
+          .slice(i)
+          .match(
+            /^<(\d+|[a-zA-Z][a-zA-Z0-9]*)(\s+[\w\-]+(?:=(?:"[^"]*"|'[^']*'|[^\s>]+))?)*\s*(\/)?>/,
+          );
         if (openingMatch) {
           const tagName = openingMatch[1];
           // Valid if it's a number or in our valid names list
--- a/test/trans.render.object.spec.jsx
+++ b/test/trans.render.object.spec.jsx
@@ -368,3 +368,27 @@ describe('trans with less-than in mathematical expression', () => {
     `);
   });
 });
+
+describe('trans with numbered tags with attributes', () => {
+  function TestComponent() {
+    return (
+      <Trans
+        defaults='Result should be a clickable link <0 href="https://www.google.com">Google</0>'
+        components={[<a>dummy</a>]}
+      />
+    );
+  }
+  it('should render tags with attributes correctly', () => {
+    const { container } = render(<TestComponent />);
+    expect(container.firstChild).toMatchInlineSnapshot(`
+      <div>
+        Result should be a clickable link 
+        <a
+          href="https://www.google.com"
+        >
+          Google
+        </a>
+      </div>
+    `);
+  });
+});
--- a/test/trans.render.spec.jsx
+++ b/test/trans.render.spec.jsx
@@ -186,9 +186,7 @@ describe('trans simple with custom html tag', () => {
         <strong>
           Go
         </strong>
-        &lt;video /&gt;
-        &lt;script&gt;console.warn("test")&lt;/script&gt;
-         there.
+        &lt;video /&gt;&lt;script&gt;console.warn("test")&lt;/script&gt; there.
       </div>
     `);
   });


adrai:
thx, it's included in v16.2.2

christophemenager:
I am not sure to understand the issue fixed with this PR but it breaks my usage:

```
  "noGameCodeModal": {
    "subtitle1": "It is not possible to obtain a code from our mobile application, but only from our website: <website-link>mywebsite.fr</website-link>",
  },
```

```
  <Trans
            defaults={t('noGameCodeModal.subtitle1')}
            components={{
              'website-link': (
                <Text
                  textDecorationLine="underline"
                  variant="body1_bold"
                  color="secondary"
                  onPress={() => {
                    Linking.openURL('myWebsite.fr');
                  }}
                />
              ),
            }}
          />
```

Is there something I was previously doing wrong? If not, maybe this change should be at least minor or major as it's breaking

christophemenager:
@adrai let met know if you want me to open an new issue

adrai:
@codomposer can you please have a look at this, else we'll need to revert your PR... (I suspect, to fix this properly we need [this](https://github.com/i18next/react-i18next/issues/1734#issuecomment-2681777802) to land)

codomposer:
@christophemenager 
can you elaborate more on your issue? any logs or expected behavior

christophemenager:
| Expected behavior (without this PR)    |Current behavior (with this PR |
| -------- | ------- |
|IMG_0071|IMG_F967CA6A82DD-1|

adrai:
https://react.i18next.com/latest/trans-component#alternative-usage-which-lists-the-components-v11.6.0

adrai:
@christophemenager maybe you can create an appropriate test here? https://github.com/i18next/react-i18next/blob/master/test/trans.render.object.spec.jsx

codomposer:
I am on it, will push a hotfix

codomposer:
@adrai  @christophemenager 
https://github.com/i18next/react-i18next/pull/1882
hope this should fix the issue

adrai:
@christophemenager please try with v16.2.3

christophemenager:
v16.2.3 fixes the issue, nice job!

rothsandro:
If a (user-generated) value contains "<" it will be rendered as `&lt;`. Is that the intended behavior?

```tsx
<Trans
  i18nKey="someKey"
  components={{ ... }}
  values={{ name: "Foo < Bar" /* would be a user-generated value */ }}
/>

PR #1882, opened 2025-10-30
fix hyphened component break issue

### LINKED ISSUE #1435: Unescape soft-hypen

### LINKED ISSUE #1734: Less than sign `<` breaks tag parsing with Trans component

### LINKED ISSUE #1802: Overriding React component props
## Documentation issue

I have an issue with the `Trans` Component [Overriding React component props ](https://react.i18next.com/latest/trans-component#overriding-react-component-props-v11.5.0) behaviour. I checked the `Trans` tests, and there is no test for Overwriting props; once I add one, it's failing. Is it a feature removed from the code but left in the documentation?

#### For

```js
/**
 * With messages:
 * myKey: 'This is a <CustomLink href="https://example.com/en-EN">link to example.com</CustomLink>.'
 **/  

<Trans
  i18nKey="myKey"
  components={{ 
    CustomLink: <a href="value-to-be-overridden"/> 
    }}
/>
```

#### Existing behaviour

```html
This is a <a href="value-to-be-overridden">link to example.com</a>.
```

#### Expected behaviour

```html
This is a <a href="https://example.com/en-EN">link to example.com</a>.
```

## Motivation

I want to fix the flows in the documentation and, at best, bring back the props substitution. The use case with links seems valuable, especially when using Routing. However, I can see additional ones, such as using icons as a component.

codomposer:
## Fixes

Closes #1734 
Closes #1802 
Closes #1435 

#### Checklist

- [x] tests are included

### DIFF
--- a/src/TransWithoutContext.js
+++ b/src/TransWithoutContext.js
@@ -144,7 +144,7 @@ const escapeLiteralLessThan = (str, keepArray = [], knownComponentsMap = {}) => 
       let isValidTag = false;
 
       // Check for closing tag: </number> or </name>
-      const closingMatch = str.slice(i).match(/^<\/(\d+|[a-zA-Z][a-zA-Z0-9]*)>/);
+      const closingMatch = str.slice(i).match(/^<\/(\d+|[a-zA-Z][a-zA-Z0-9-]*)>/);
       if (closingMatch) {
         const tagName = closingMatch[1];
         // Valid if it's a number or in our valid names list
@@ -163,7 +163,7 @@ const escapeLiteralLessThan = (str, keepArray = [], knownComponentsMap = {}) => 
         const openingMatch = str
           .slice(i)
           .match(
-            /^<(\d+|[a-zA-Z][a-zA-Z0-9]*)(\s+[\w-]+(?:=(?:"[^"]*"|'[^']*'|[^\s>]+))?)*\s*(\/)?>/,
+            /^<(\d+|[a-zA-Z][a-zA-Z0-9-]*)(\s+[\w-]+(?:=(?:"[^"]*"|'[^']*'|[^\s>]+))?)*\s*(\/)?>/,
           );
         if (openingMatch) {
           const tagName = openingMatch[1];
--- a/test/trans.render.object.spec.jsx
+++ b/test/trans.render.object.spec.jsx
@@ -392,3 +392,27 @@ describe('trans with numbered tags with attributes', () => {
     `);
   });
 });
+
+describe('trans with hyphenated component names', () => {
+  function TestComponent() {
+    return (
+      <Trans
+        defaults="Visit our website: <website-link>mywebsite.fr</website-link>"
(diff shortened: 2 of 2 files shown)


adrai:
thx @codomposer

PR #1898, opened 2026-01-05
Fix/no suspense

### LINKED ISSUE #1896: When not using suspense, provoke a typescript error if `t` is accessed without first checking `ready` (optional setting)
## 🚀 Feature Proposal

Add a new boolean type option that, when enabled and suspense is disabled, makes the UseTranslationResponse type evaluate to something like `{
  t: TFunction;
  ready: true;
} | {
  ready?: false;
}`.

That would provoke a type error if `t` is accessed without first checking `ready`.

## Motivation

Disabling suspense can result in snappier loading behavior, but if you forget to check `ready`, you tend to get i18n keys visibly rendered on your site.
Unfortunately, to make use of this you would have to avoid destructuring the `useTranslation` return value and use `i18n.ready` and `i18n.t`, which is why it should be opt-in behavior.

I'm also not sure if and how it would affect the non-hook stuff.

## Example

```
declare module 'i18next' {
  interface CustomTypeOptions {
    strictNoSuspense: true,
  }
}
const Component = () => {
  const i18n = useTranslation(['translation'], {useSuspense: false});
  i18n.t($ => $.foo); // error!
  i18n.ready && i18n.t($ => $.foo); // ok
};
```

eureka0928:
## Add `strictNoSuspense` TypeScript option for safer `useTranslation` usage

Fixes #1896

### Description

When using `useSuspense: false`, forgetting to check `ready` before accessing `t` causes i18n keys to render on the page. This PR adds an optional TypeScript setting that enforces checking `ready` before using `t`.

### Usage

```typescript
// Enable in your i18next.d.ts
declare module 'i18next' {
  interface CustomTypeOptions {
    strictNoSuspense: true;
  }
}
```

```typescript
const result = useTranslation('ns', { useSuspense: false });

result.t('key');  // TypeScript error: t is possibly undefined

if (result.ready) {
  result.t('key');  // OK
}

result.ready && result.t('key');  // Also OK
```

When `useSuspense: true` (default) or when `strictNoSuspense` is not enabled, behavior remains unchanged.

### Changes

- Added `UseTranslationResponseNotReady`, `UseTranslationResponseReady`, and `UseTranslationResponseStrict` types to `index.d.ts`
- Added overloads to `useTranslation` that return discriminated union when `useSuspense: false` and `strictNoSuspense: true`
- Added test suite in `test/typescript/strict-no-suspense/`

#### Checklist

- [ ] tests are included

Contribution by Gittensor, see my contribution statistics at https://gittensor.io/miners/details?githubId=42954461

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -190,6 +190,8 @@ export function useSSR(initialI18nStore: Resource, initialLanguage: string): voi
 // If the version is earlier than i18next v25.4.0, enableSelector does not exist in TypeOptions, so a conditional type is used to maintain type compatibility.
 type _EnableSelector = TypeOptions extends { enableSelector: infer U } ? U : false;
 
+type _StrictNoSuspense = TypeOptions extends { strictNoSuspense: infer U } ? U : false;
+
 export type UseTranslationOptions<KPrefix> = {
   i18n?: i18n;
   useSuspense?: boolean;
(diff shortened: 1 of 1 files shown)


adrai:
if this lands, we need to update the docs, either [here](https://www.i18next.com/overview/typescript) or [here](https://react.i18next.com/latest/usetranslation-hook#not-using-suspense) or both.

fury0928:
> if this lands, we need to update the docs, either [here](https://www.i18next.com/overview/typescript) or [here](https://react.i18next.com/latest/usetranslation-hook#not-using-suspense) or both.

@adrai do i need to update the docs as well?

adrai:
> > if this lands, we need to update the docs, either [here](https://www.i18next.com/overview/typescript) or [here](https://react.i18next.com/latest/usetranslation-hook#not-using-suspense) or both.
> 
> @adrai do i need to update the docs as well?

if this PR lands, yes - that would be nice

fury0928:
> > > if this lands, we need to update the docs, either [here](https://www.i18next.com/overview/typescript) or [here](https://react.i18next.com/latest/usetranslation-hook#not-using-suspense) or both.
> > 
> > 
> > @adrai do i need to update the docs as well?
> 
> if this PR lands, yes - that would be nice

@adrai i will put the PR for docs once this is landed.
When would you review and this can land?

adrai:
> When would you review and this can land?

depends on @marcalexiei or @ahrjarrett

PR #1900, opened 2026-01-10
fix #1899

### LINKED ISSUE #1899: Type errors when you've declared a resources type
## 🐛 Bug Report

After upgrading to react-i18next ≥16.1.1, I now get the following type errors in my project:

```
node_modules/react-i18next/index.d.ts:50:46 - error TS2344: Type 'string' does not satisfy the constraint 'FilterKeysByContext<ParseKeysByKeyPrefix<KeysByTOptions<TOpt>[$FirstNamespace<NsByTOptions<Ns, TOpt>>], KPrefix> | ParseKeysByNamespaces<...>, TOpt["context"]>'.

50   Key extends ParseKeys<Ns, TOpt, KPrefix> = string,
                                                ~~~~~~

node_modules/react-i18next/index.d.ts:76:46 - error TS2344: Type 'string' does not satisfy the constraint 'FilterKeysByContext<ParseKeysByKeyPrefix<KeysByTOptions<TOpt>[$FirstNamespace<NsByTOptions<Ns, TOpt>>], KPrefix> | ParseKeysByNamespaces<...>, TOpt["context"]>'.

76   Key extends ParseKeys<Ns, TOpt, KPrefix> = string,
                                                ~~~~~~

node_modules/react-i18next/index.d.ts:99:48 - error TS2344: Type 'string' does not satisfy the constraint 'FilterKeysByContext<ParseKeysByKeyPrefix<KeysByTOptions<TOpt>[$FirstNamespace<NsByTOptions<Ns, TOpt>>], KPrefix> | ParseKeysByNamespaces<...>, TOpt["context"]>'.

99     Key extends ParseKeys<Ns, TOpt, KPrefix> = string,
                                                  ~~~~~~

node_modules/react-i18next/index.d.ts:126:48 - error TS2344: Type 'string' does not satisfy the constraint 'FilterKeysByContext<ParseKeysByKeyPrefix<KeysByTOptions<TOpt>[$FirstNamespace<NsByTOptions<Ns, TOpt>>], KPrefix> | ParseKeysByNamespaces<...>, TOpt["context"]>'.

126     Key extends ParseKeys<Ns, TOpt, KPrefix> = string,
                                                   ~~~~~~
```

## To Reproduce

You must declare a `resources` type in your project, like so:

```typescript
declare module "i18next" {
  interface CustomTypeOptions {
    defaultNS: "app";
    keySeparator: ".";
    resources: {
      app: { foo: string; bar: string };
    };
  }
}
```

This causes the above errors because, just as TypeScript says, `string` is no longer a valid default value for the IcuTrans type parameters. It now accepts only the specific strings "app.foo" or "app.bar". By eyeballing it I think that the default should be `ParseKeys<Ns, TOpt, KPrefix>`, maybe?

## Expected behavior

Declaring a `resources` type for type-safe invocation of the `t` function should be fine, and not create type errors elsewhere inside my i18next node modules.

## Your Environment

- *runtime version*: Node 22.20.0
- *i18next version*: 25.7.3
- *os*: NixOS 25.11

cellog:
This change fixes the issue raised by @robintown. The types for `Trans` did not have a default value for `Key`, which is what was causing the breakage.

Basically, IcuTrans requires a key, so there is always a value to infer from. For Trans, where the key is optional, the value was inferred as the full `ParseKeys<...>` type. The bug was introducing a default of `= string`. This is unnecessary and was the root cause of the problem.

@robintown can you please try this fix on your local codebase and confirm it works?

#### Checklist

- [x] tests are included

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -47,7 +47,7 @@ export interface IcuTransContentDeclaration {
  * Props for IcuTransWithoutContext component (no React context)
  */
 export type IcuTransWithoutContextProps<
-  Key extends ParseKeys<Ns, TOpt, KPrefix> = string,
+  Key extends ParseKeys<Ns, TOpt, KPrefix>,
   Ns extends Namespace = _DefaultNamespace,
   KPrefix = undefined,
   TContext extends string | undefined = undefined,
@@ -73,7 +73,7 @@ export type IcuTransWithoutContextProps<
  * Props for IcuTrans component (with React context support)
  */
 export type IcuTransProps<
-  Key extends ParseKeys<Ns, TOpt, KPrefix> = string,
+  Key extends ParseKeys<Ns, TOpt, KPrefix>,
   Ns extends Namespace = _DefaultNamespace,
   KPrefix = undefined,
   TContext extends string | undefined = undefined,
@@ -96,7 +96,7 @@ export type IcuTransProps<
  */
 export interface IcuTransComponent {
   <
-    Key extends ParseKeys<Ns, TOpt, KPrefix> = string,
+    Key extends ParseKeys<Ns, TOpt, KPrefix>,
     Ns extends Namespace = _DefaultNamespace,
     KPrefix = undefined,
     TContext extends string | undefined = undefined,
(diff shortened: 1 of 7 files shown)


adrai:
@cellog fyi: the typescript tests fails

cellog:
I gotchu

PR #1916, opened 2026-04-17
Security: Global singleton i18n instance can leak state across SSR requests

tuanaiseo:
## Problem

The library stores the active i18n instance in a module-level variable (`i18nInstance`). In server-side rendering environments, module globals are shared across concurrent requests. If different requests/users initialize different instances or language/resource state, this can cause cross-request state bleed (wrong language, namespace data, or tenant-specific resources being reused). `getInitialProps` in `src/context.js` reads from this shared singleton, which amplifies this risk.

**Severity**: `medium`
**File**: `src/i18nInstance.js`

## Solution

Avoid process-global mutable i18n state for SSR. Pass i18n through request-scoped context/providers only, or keep a per-request instance map keyed by request context. Document that singleton APIs are client-only and provide SSR-safe alternatives.

## Changes

- `src/i18nInstance.js` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

### DIFF
--- a/src/i18nInstance.js
+++ b/src/i18nInstance.js
@@ -1,7 +1,8 @@
 let i18nInstance;
+const isBrowser = typeof window !== 'undefined';
 
 export const setI18n = (instance) => {
-  i18nInstance = instance;
+  if (isBrowser) i18nInstance = instance;
 };
 
-export const getI18n = () => i18nInstance;
+export const getI18n = () => (isBrowser ? i18nInstance : undefined);


adrai:
Thanks for the submission, but we're going to close this.

To be candid: the framing and structure of this PR (and #1917) read as AI-generated with minimal human review — formulaic "Severity" labels, unchecked test boxes, and a "fix" that doesn't actually address the stated concern. Please don't open security-framed PRs without verifying the change against the codebase.

On the technical side, the proposed change makes `setI18n`/`getI18n` no-ops on the server, which actively breaks SSR rather than fixing it:

- `getInitialProps` / `composeInitialProps` in `src/context.js` are explicitly designed to run on the server and rely on `getI18n()`. With this patch they would always warn and return `{}`, silently breaking data hydration.
- `Trans`, `useTranslation`, `IcuTrans` all fall back to the global during server render when no prop/context instance is provided. This fallback would become `undefined` on the server, breaking any SSR setup that doesn't wrap with `I18nextProvider`.

For true SSR isolation the documented pattern already exists: create a per-request instance (typically via `i18next-http-middleware`, which clones per request) and pass it through `<I18nextProvider i18n={req.i18n}>`, as shown in the [SSR docs](https://react.i18next.com/latest/ssr) and [`example/razzle-ssr`](https://github.com/i18next/react-i18next/tree/master/example/razzle-ssr). The client-side `useSSR` hook complements this by hydrating from `initialI18nStore` and explicitly opts out for cloned instances. The module-level singleton is a client-side convenience — not a shared mutable request store.

Closing.

PR #1917, opened 2026-04-17
Security: Global mutable default options are shared across consumers/requests

tuanaiseo:
## Problem

`defaultOptions` is a mutable module-global object updated via `setDefaults`. In SSR or multi-instance usage, one initialization path can overwrite defaults for others, creating cross-request/config contamination. While mostly integrity-related, it can affect security-sensitive rendering behavior (e.g., escaping/unescaping behavior) if options differ by context.

**Severity**: `low`
**File**: `src/defaults.js`

## Solution

Use immutable/request-scoped configuration instead of mutating module globals. For SSR, derive options from the active i18n instance/context at render time. If globals must remain, clearly document non-thread-safe behavior.

## Changes

- `src/defaults.js` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

### DIFF
--- a/src/defaults.js
+++ b/src/defaults.js
@@ -15,7 +15,16 @@ let defaultOptions = {
 };
 
 export const setDefaults = (options = {}) => {
-  defaultOptions = { ...defaultOptions, ...options };
+  defaultOptions = {
+    ...defaultOptions,
+    ...options,
+    ...(options.transKeepBasicHtmlNodesFor
+      ? { transKeepBasicHtmlNodesFor: [...options.transKeepBasicHtmlNodesFor] }
+      : {}),
+  };
 };
 
-export const getDefaults = () => defaultOptions;
+export const getDefaults = () => ({
+  ...defaultOptions,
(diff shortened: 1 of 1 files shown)


adrai:
Thanks, but we're going to close this.

To be candid: this PR (along with #1916) reads as AI-generated with minimal human review — formulaic severity framing, unchecked test boxes, and a "security" justification that doesn't hold up when you trace the code. Please don't open security-framed PRs without verifying the change.

On the technical side:

- `getDefaults()` is called inside `useTranslation` and `TransWithoutContext` on every render. Returning a fresh object + cloned array on each call produces unnecessary allocations on a hot path.
- It doesn't actually prevent cross-request contamination — `defaultOptions` is still a single module-level mutable object, so concurrent `setDefaults` calls still overwrite each other. This is cosmetic immutability, not request isolation.
- For true SSR isolation, configuration should come from the per-request i18n instance (`i18n.options.react`), which is already merged in at the call sites. That is the recommended pattern.

Closing.

PR #1921, opened 2026-05-14
fix(types): accept selector i18nKey on <Trans> under enableSelector: 'strict'

Faithfinder:
## Problem

Under `CustomTypeOptions.enableSelector: 'strict'`, every `<Trans i18nKey={($) => ...}>` call site fails to typecheck:

```
Type '($: any) => any' is not assignable to type '"some-key-from-resources" | ... | undefined'.
Parameter '$' implicitly has an 'any' type.
```

The cause is in `TransWithoutContext.d.ts`:

```ts
export const Trans: _EnableSelector extends true | 'optimize' ? TransSelector : TransLegacy;
```

`'strict'` doesn't match `true | 'optimize'`, so `Trans` collapses to `TransLegacy`, whose `i18nKey: Key` is a string union. The runtime, however, accepts selector `i18nKey` regardless of mode — `TransWithoutContext.js` calls `keyFromSelector(i18nKey)` whenever `typeof i18nKey === 'function'`. So this is a type-side oversight; the runtime is already correct.

## Repro

1. Set `CustomTypeOptions.enableSelector: 'strict'`.
2. `<Trans i18nKey={($) => $.someNs["Some key"]}>` — errors at compile time, renders fine at runtime.

## Fix

Widen the conditional to include `'strict'`:

```diff
-export const Trans: _EnableSelector extends true | 'optimize' ? TransSelector : TransLegacy;
+export const Trans: _EnableSelector extends true | 'optimize' | 'strict' ? TransSelector : TransLegacy;
```

## Tests

Adds `test/typescript/selector-strict/` mirroring `selector-optimize/` (same `i18next.d.ts` / `tsconfig.json` shape, with `enableSelector: 'strict'`). The `Trans.test.tsx` covers the strict-mode selector contract:

- Default namespace requires an explicit prefix (`$.custom.foo`).
- Flat-primary paths (`$.foo`) raise a TypeError.
- Named-namespace + array-namespace paths still typecheck via the namespace prefix.
- `t` from `useTranslation` flows through.
- Interpolation cases mirror the optimize-mode tests.

`keyPrefix` cases are intentionally not covered — strict mode interacts with `keyPrefix` differently and is out of scope here. All 383 type-tests in the existing suite still pass.

```
$ npx vitest --config vitest.workspace.typescript.mts run
 Test Files  42 passed (42)
      Tests  383 passed (383)
Type Errors  no errors
```

### DIFF
--- a/TransWithoutContext.d.ts
+++ b/TransWithoutContext.d.ts
@@ -106,7 +106,9 @@ export interface TransSelector {
   ): React.ReactElement;
 }
 
-export const Trans: _EnableSelector extends true | 'optimize' ? TransSelector : TransLegacy;
+export const Trans: _EnableSelector extends true | 'optimize' | 'strict'
+  ? TransSelector
+  : TransLegacy;
 
 export function nodesToString(
   children: React.ReactNode,
--- a/test/typescript/selector-strict/Trans.test.tsx
+++ b/test/typescript/selector-strict/Trans.test.tsx
@@ -1,0 +1,109 @@
+import { describe, it, expectTypeOf } from 'vitest';
+import * as React from 'react';
+import { Trans, useTranslation } from 'react-i18next';
+
+describe('<Trans /> under enableSelector: "strict"', () => {
+  describe('default namespace', () => {
+    it('requires an explicit namespace prefix', () => {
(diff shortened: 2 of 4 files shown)


adrai:
Shipped in **react-i18next 17.0.8**. Thanks @Faithfinder!

PR #1930, opened 2026-09-01
fix(types): allow selector keyPrefix in useTranslation under enableSelector 'strict'

hovelopin:
## Problem

`useTranslation` is gated on `_EnableSelector extends true | 'optimize'`, which
excludes `'strict'`. Under `enableSelector: 'strict'` it resolves to `UseTranslationLegacy`, so the selector-form `keyPrefix` overload is gone entirely:

```ts
// test/typescript/selector-strict — enableSelector: 'strict'
const { t } = useTranslation('alternate', {
  keyPrefix: ($) => $.alternate.foobar.deep,
});
//           ^ Type '($: any) => any' is not assignable to type 'undefined'.
```

`Trans` in `TransWithoutContext.d.ts` already gates on `true | 'optimize' | 'strict'`,
so `useTranslation` is the one export `'strict'` was never added to.

Companion to i18next/i18next#2446, which fixed the same omission in `getFixedT`.

## Fix

```diff
-export const useTranslation: _EnableSelector extends true | 'optimize'
+export const useTranslation: _EnableSelector extends true | 'optimize' | 'strict'
   ? UseTranslationSelector
   : UseTranslationLegacy;
```

Adds `useTranslation.test.ts` to the existing `selector-strict` scenario. It was the
only scenario without one, which is why this went unnoticed:

```ts
it('should work with a selector keyPrefix', () => {
  const [t] = useTranslation('alternate', {
    keyPrefix: ($) => $.alternate.foobar.deep,
  });

  expectTypeOf(t(($) => $.deeper.deeeeeper)).toEqualTypeOf<'foobar'>();
});
```

Reverting the one-line change fails 3 of its cases.

#### Checklist

- [x] tests are included

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -222,7 +222,7 @@ export type FallbackNs<Ns> = Ns extends undefined
     ? Ns
     : _DefaultNamespace;
 
-export const useTranslation: _EnableSelector extends true | 'optimize'
+export const useTranslation: _EnableSelector extends true | 'optimize' | 'strict'
   ? UseTranslationSelector
   : UseTranslationLegacy;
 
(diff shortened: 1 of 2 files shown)


adrai:
thx, included in v17.0.13

